PR #355, opened 2020-03-18
test-suite: attempt to sub-parallelize between jobs

### LINKED ISSUE #354: test-suite: parallelize .travis.yml to avoid idle CPU and very long test build times
Currently, the average longest time for the longest running test is about 1hr 30min, but over 6 parallel jobs allowed by TravisCI. With 43 tests, this could take 3 hours to run.

Travis CI has no mechanism to speed up tests (that I know of), but perhaps what we can do is further divide up those tests and parallelize them over the number of CPUs present so that they'll get scheduled piecewise but at least let's try to reduce the overall time spent trying to test.

Here is a Go program I've written to parallelize tests on the same machine.
```go
package main

import (
	"context"
	"errors"
	"math/rand"
	"os"
	"os/exec"
	"runtime"
	"strings"
	"sync"

	"github.com/orijtech/otils"
)

func main() {
	testApps := otils.NonEmptyStrings(strings.Split(os.Getenv("DJANGO_TEST_APPS"), " ")...)
	if len(testApps) == 0 {
		panic("No DJANGO_TEST_APPS passed in")
	}

	rand.Shuffle(len(testApps), func(i, j int) {
		testApps[i], testApps[j] = testApps[j], testApps[i]
	})

	// Otherwise, we'll parallelize the builds.
	nParallel := runtime.GOMAXPROCS(0)
	println(nParallel)

	shutdownCtx, cancel := context.WithCancel(context.Background())
	defer cancel()

	var wg sync.WaitGroup
	defer wg.Wait()

	stride := 1 + (len(testApps) / nParallel)
	for i := 0; i < len(testApps); i += stride {
		apps := testApps[i : i+stride]
		wg.Add(1)
		go runTests(shutdownCtx, &wg, apps, "django_test_suite.sh")
	}
}

func runTests(ctx context.Context, wg *sync.WaitGroup, djangoApps []string, testSuiteScriptPath string) error {
	defer wg.Done()
	defer println("Done here")

	if len(djangoApps) == 0 {
		return errors.New("Expected at least one app")
	}

	cmd := exec.CommandContext(ctx, "bash", testSuiteScriptPath)
	cmd.Env = append(os.Environ(), `DJANGO_TEST_APPS=`+strings.Join(djangoApps, " ")+``)
	cmd.Stderr = os.Stderr
	cmd.Stdout = os.Stdout
	if err := cmd.Run(); err != nil {
		panic(err)
	}
}
```

and with it, we could compile a binary and check it into source and then reduce on the number of .travis.yml entries.

odeke-em:
Between TravisCI jobs which can only concurrently run
6 jobs at a time, we can perhaps take advantage of
parallelism and not leaving CPUs idle but scheduling
parallel DJANGO_TEST_APPS.

Fixes #354

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,56 +7,29 @@ env:
         - GOOGLE_APPLICATION_CREDENTIALS=~/creds.json
         - RUNNING_SPANNER_BACKEND_TESTS=1
     jobs:
-        - DJANGO_TEST_APPS="admin_changelist admin_custom_urls"
-        - DJANGO_TEST_APPS="admin_docs"
-        - DJANGO_TEST_APPS="admin_inlines admin_ordering"
-        - DJANGO_TEST_APPS="aggregation"
-        - DJANGO_TEST_APPS="aggregation_regress annotations"
+        - DJANGO_TEST_APPS="admin_changelist admin_custom_urls admin_docs admin_inlines"
+        - DJANGO_TEST_APPS="admin_ordering aggregation aggregation_regress annotations backends"
         # Commented out because they take longer 2hr and TravisCI unconditionally terminates them.
         # - DJANGO_TEST_APPS="auth_tests"
-        - DJANGO_TEST_APPS="backends basic"
-        - DJANGO_TEST_APPS="bulk_create choices custom_columns"
-        - DJANGO_TEST_APPS="cache"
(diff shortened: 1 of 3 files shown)


odeke-em:
No longer relevant since we are moving to using Kokoro on Google Cloud.

Issue #289, opened 2020-02-19
tests: move builds from Travis to Kokoro on Google Cloud Build


odeke-em:
Let's look at moving to Kokoro which we can get on Google Cloud infrastructure. Travis CI is slow, pricey (costs me about $400/month), throttled.

odeke-em:
Alright, now that this is here, I had pinged @JustinBeckwith offline and he let me know that he can help out with moving to Kokoro.

## Setup
The script that runs tests is `./django_test_suite.sh`

## Environment variables
Variable|Comment
---|---
DJANGO_TEST_APPS|The various Django test apps to use as in the .travis.yml file. For example DJANGO_TEST_APPS="validation view_tests"
GOOGLE_APPLICATION_CREDENTIALS|The application default credentials

PR #406, opened 2020-04-02
.kokoro: add parallel builds/workers

odeke-em:
Phase 4 of getting CI to work with Kokoro:
* Add django test apps into a single file, sorted alphabetically
* Adjust the per machine parallelizer to accept a job index ID
and count and then grab the chunk of tests for it

Updates #289

.kokoro: attempt to parallelize tests on multiple cores

install Django test requirements

Following the test failure from
* https://source.cloud.google.com/results/invocations/ffc94ee3-2ae0-4aa9-b3dc-8d85dd602960/log
in which we had skipped installing Django requirements.

build.sh: add -y to apt-get install


odeke-em:
@busunkim96 please take a look at the 8 worker_*.cfg files added to .kokoro/presubmit/ and .kokoro/continuous/
Thank you!

parallelize_tests.go
```diff
@@ -9,19 +9,44 @@ package main
...
+	"log"
+	"math"
 	"math/rand"
 	"os"
 	"os/exec"
 	"os/signal"
 	"runtime"
+	"strconv"
 	"strings"
 	"sync"
```
busunkim96, line 22: Would you mind rewriting this script in Python? It will make this repository easier to maintain for other folks  who may not be familiar with Go.

odeke-em:
Perhaps a little later after we’ve delivered the project past GA, because:
a) that code implicitly relies on the underlying scheduler to run each
parallel routine on a separate core — doing so in Python requires a whole
lot of work and am not sure we’d easily achieve the same result easily

b) not having ported tests running has unfortunately set us back by about
2.5 days, yet unfortunately we are quite time strapped

For now there is a checked in binary.

.kokoro/continuous/worker_0.cfg
```diff
@@ -0,0 +1,4 @@
+env_vars: {
+    key: "DJANGO_WORKER_INDEX"
+    value: 0
```
busunkim96, line 3: Ah it looks like these need to be strings
```suggestion
    value: "0"
```

busunkim96:
👍 Sounds good. It looks like the `env_var` values need to be strings.

.kokoro/continuous/worker_0.cfg
odeke-em: Ack, thank you! Let me make that update!

odeke-em:
>👍 Sounds good. It looks like the env_var values need to be strings.

Done, thank you!

busunkim96:
Hmm, it looks like the jobs are now failing with quota exhausted errors. Have you run into this in the past?

```
google.api_core.exceptions.ResourceExhausted: 429 Quota exceeded for quota metric 'Administrative requests' and limit 'Administrative requests per minute' of service 'spanner.googleapis.com'
```

odeke-em:
Thank you for the report @busunkim96!
Indeed, seems like Cloud Spanner’s very low administrative RPC limits for DDL are at play here.
They have a limit of 5QPS moreover averaged over 100 seconds per https://cloud.google.com/spanner/quotas#administrative_limits
9CAE5859-6570-4BC3-8DDE-0FDD8A30722F

This parallelization does perhaps 32QPS (8 workers * 4 cores/worker) sustained. Perhaps I’ll introduce some throttling of sorts to keep below the very low threshiods for Cloud Spanner. If you have the ability to bump up quotas, that’d be much appreciated and I’ll kindly loop in @vmanghnani @skuruppu too about helping with increasing quotas.

The other trick would be to have maybe 4 unique projects or use projects from a pool.

PR #410, opened 2020-04-04
tests: use Spanner emulator

odeke-em:
tests: use Spanner emulator

Uses the just released Cloud Spanner emulator
along with the python-spanner libraries that
recognize SPANNER_EMULATOR_HOST in the environment.

In preliminary tests on DJANGO_TEST_APPS=basic,
the running time was cut down to ~30sec using the emulator
instead of 8min when making calls to Cloud Spanner production.

Given that the emulator does support some features listed in
https://github.com/GoogleCloudPlatform/cloud-spanner-emulator#features-and-limitations

we'll need to keep building the list of unsupported features
as failing tests as revealed.

To generate tests to skip that can't yet the Spanner emulator's
features, I downloaded all the logs for the 8 workers then ran
```shell
find ~/Downloads/ -type f -name 'kokoro*' -exec grep 'ERROR:' {} \; | cut -d":" -f2 | awk '{ print $2"."$1 }' | sed s'/(//g' | sed s'/)//g' | sort | uniq | while read F;do echo "            '$F',  # noqa";done >> django_spanner/features.py
```


odeke-em:
Using the Spanner emulator seems to not be feasible for our use case since it can't support a commonly used pattern: Unresolved parameter types when None is used even as documented in https://github.com/GoogleCloudPlatform/cloud-spanner-emulator#features-and-limitations
Screen Shot 2020-04-03 at 11 35 52 PM


and the impact would involve skipping thousands of individual tests. I am just on two worker failure logs:
a) [worker_0/log](https://source.cloud.google.com/results/invocations/7ae3c117-3f63-47d7-8087-916e7e8a95b3/targets/cloud-devrel%2Fclient-libraries%2Fpython%2Fgoogleapis%2Fpython-spanner-django%2Fpresubmit%2Fworker_0/log)
b) [worker_4/log](https://source.cloud.google.com/results/invocations/f6ee43d2-0a7a-401c-9f7f-06cfae5c95af/targets/cloud-devrel%2Fclient-libraries%2Fpython%2Fgoogleapis%2Fpython-spanner-django%2Fpresubmit%2Fworker_4/log)

and on running a script to automatically create tests to skip, it gives 578
```shell
$ find ~/Downloads/ -type f -name 'kokoro*' -exec grep 'ERROR:' {} \; | cut -d":" -f2 | tr "()" "-" | sed s/-//g | awk '{ print $2.$1 }' | sort | uniq  | wc -l
     578
```
and that's while we are at 1/4 of the workers; then imagine how many tests we'd have to skip?

odeke-em:
We can turn off the emulator when running integration tests. For now this will work and can parallelize tests properly.

PR #414, opened 2020-04-05
parallelize_tests: create a unique instance per worker

### LINKED ISSUE #413: ci: lots of distracting flakes from a very low administrative quota limit for Cloud Spanner
While testing we are encountering a whole lot of flakes in concurrent builds. Unfortunately that might come from us using the same Cloud Spanner instance instance.

The newly released Cloud Spanner emulator is unusable for this project because:
a) It doesn't support INFORMATION_SCHEMA.CONSTRAINT_USAGE
b) For parameters whose values are None, their parameter types will be unspecified
and the emulator will throw errors

I spent a whole day trying to fix this with various tricks but really the problem here is that we are using a single instance. I think as one of the last attempts, we can try creating an instance per test and then tear it down on exiting.

odeke-em:
Ensures that each worker creates a unique Spanner instance
which will be DELETED as the worker is exiting!

Potentially solves the issue with very low administrative
quotas for Cloud Spanner, since now each worker will have
its own limits instead of trying to use the exact same instance.

If this works, we'll be able to massively parallelize our tests!

Fixes #413.

### DIFF
--- a/.kokoro/build.sh
+++ b/.kokoro/build.sh
@@ -62,7 +62,7 @@ then
     # Not using the emulator!
     # Hardcode the max number of workers since Spanner has a very low
     # QPS for administrative RPCs of 5QPS (averaged every 100 seconds)
-    export DJANGO_WORKER_COUNT=4
+    export DJANGO_WORKER_COUNT=5
 else
     export DJANGO_WORKER_COUNT=$(ls .kokoro/presubmit/worker* | wc -l)
     # Install and start the Spanner emulator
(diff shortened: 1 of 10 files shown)


odeke-em:
Seems like there isn't anything else to do besides requesting for a quota bump to the highest tier. I've sent an email to some Googlers to help bump this up. When they get back, we'll perhaps try to run this again.

odeke-em:
Attempting to re-run this CI job.

odeke-em:
I just realized that given that we have both presubmit and continuous checks running, and they use identical configurations! This means that our quota is running out even faster if check `continuous` is also running in parallel. We should perhaps only have a single worker for `continuous`.

odeke-em:
@busunkim96 please help me cancel the current presubmit and continuous checks that are running. I want to make `continuous` only run one worker. Maybe this might help with quota problems. I am going to update this code to remove all the workers but one for the `continuous` check

PR #417, opened 2020-04-07
ci: install Django before django-spanner

timgraham:
Otherwise, tests crash due to importing django in django_spanner/__init__.py
and importing django_spanner in setup.py.

Regression in 3a244267ffc6cabe847511896a0616fa161d3bfe.

### DIFF
--- a/.kokoro/build.sh
+++ b/.kokoro/build.sh
@@ -33,7 +33,7 @@ export PROJECT_ID=$(cat "${KOKORO_GFILE_DIR}/project-id.json")
 python3.6 -m pip install --upgrade --quiet tox flake8 isort
 python3.6 -m tox --version
 
-python3.6 -m tox
+# Run code style checks
 python3.6 -m isort --recursive --check-only --diff
 python3.6 -m flake8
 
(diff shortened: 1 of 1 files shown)


timgraham:
@odeke-em, it looks like the version you tagged yesterday won't be easily installable (unless google-cloud is already installed) because of this issue.

I guess `version=__import__('django_spanner').__version__` in `setup.py` isn't workable because `django_version/__init__.py` imports `google.api_core.datetime_helpers` which may not be installed. Perhaps we'll just have to duplicate the version number in `setup.py` and `django_version/__init__.py` (https://github.com/googleapis/python-spanner-django/pull/418) and remember to update it in both places (I would have bumped it to 2.2a1 before your release yesterday... although it doesn't matter much since we're not releasing to PyPI... 2.2a0 means "2.2 pre-alpha" while 2.2 means "2.2 alpha 1").

odeke-em:
Sheesh, that's crazy! Thanks for catching that. Oddly the CI accepted it, did you want to restore this PR or is there something else we can do?

timgraham:
I didn't try to run kokoro on the original change (https://github.com/googleapis/python-spanner-django/pull/416) (if that's what you meant by "Oddly the CI accepted it") since the build hasn't been working due to quota issues.

This attempt at fixing the issue is superseded by https://github.com/googleapis/python-spanner-django/pull/418.

PR #418, opened 2020-04-07
fix django-spanner install when django and/or google-cloud aren't installed

timgraham:
If setup.py imports django_spanner, Django and google.api_core.datetime_helpers
must be installed before django-spanner.

Regression in 3a244267ffc6cabe847511896a0616fa161d3bfe.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -14,7 +14,10 @@ install_requires = [
 
 setup(
         name='django-spanner',
-        version=__import__('django_spanner').__version__,
+        # Duplicate version here rather than using
+        # __import__('django_spanner').__version__ because that file imports
+        # django and google.cloud which may not be installed.
+        version='2.2a0',
         author='Google LLC',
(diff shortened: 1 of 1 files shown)


timgraham:
(build still broken due to unrelated errors... "panic: rpc error: code = ResourceExhausted desc = Project 1065521786570 cannot add 1 nodes in region us-west2.")

odeke-em:
>(build still broken due to unrelated errors... "panic: rpc error: code = ResourceExhausted desc = Project 1065521786570 cannot add 1 nodes in region us-west2.")

Yeah, the account attached to the CI has very low quota issues. I sent an email out to the Cloud Spanner engineering team asking for an increase but they've had to route that request as well through their product team, so we can wait and twiddle our thumbs, or perhaps just look at using CircleCI which has familiar semantics to TravisCI and will help us out, as Google has accounts for CircleCI.

PR #430, opened 2020-04-14
django: fix regex lookup with F expression crash

### LINKED ISSUE #382: django: regex lookup with an F expression crashes
#251 fixed icontains with F expressions (e.g. `Employee.objects.filter(firstname__iexact=F('lastname')`. The same issue affects regex/iregex. It looks like we'll need to add such tests to Django's test suite.

(It may appears that  contains/icontains, startswith, endswith, istartswith, and iendswith are also affected but these are the subject of #178.)

timgraham:
fixes #382


timgraham:
There isn't a test in Django's test suite for this case. I'm adding a test to the Django fork and will contribute it upstream as well.

odeke-em:
Ack, thank you!

PR #431, opened 2020-04-15
ci: use all workers for presubmit check

timgraham:


timgraham:
Should the comment be updated?
```
# Hardcode the max number of workers since Spanner has a very low
# QPS for administrative RPCs of 5QPS (averaged every 100 seconds)
```
or should we go back to using `$(ls .kokoro/presubmit/worker* | wc -l)` (unclear to me what the difference, if any, would be)

odeke-em:
Please go ahead and remove that comment entirely and let’s use all the
available workers with the $(ls ..) since our QPS got bumped up high.

timgraham:
Keep "DJANGO_WORKER_COUNT=1 # Only using 1 worker for continuous" ?

odeke-em:
Perhaps let’s keep that at 1.

timgraham:
One worker failed with `panic: rpc error: code = ResourceExhausted desc = Project 1065521786570 cannot add 1 nodes in region us-west1.`.

odeke-em:
@timgraham thanks for the report! @busunkim96 and I were just chatting about this offline and showed me the number of instances currently present
django-instances


It seems like continuous taking too long lingers those instances for ages which then causes a quota limit.
We don't even currently need continuous since it is running the exact same code as in presubmit.

Could we please perhaps set continuous worker count to 0? After that, these tests should re-run, and hopefully work alright.

timgraham:
Still an error on one worker: `Quota exceeded for quota metric 'Instance create requests' and limit 'Instance create requests per minute' of service 'spanner.googleapis.com' for consumer 'project_number:1065521786570'.`

But continuous builds are probably still running until this is merged, right?

odeke-em:
Deep sigh!! Indeed, they are still running. @busunkim96 could you please help us stop all builds for this project?

And yes @timgraham I think this PR is good to go, and after merge and after @busunkim96 has helped us out, we can then rebase the other PRs.

Thank you!

odeke-em:
Please rebase from master.

timgraham:
Still seeing "Quota exceeded for quota metric 'Instance create requests' and limit 'Instance create requests per minute' of service 'spanner.googleapis.com' for consumer 'project_number:1065521786570'."

odeke-em:
Ahh I see. In that case let’s just disable continuous, revert to keeping
presubmit at a hard coded value then move on.

PR #432, opened 2020-04-16
reorganize limitations in README

timgraham:


README.md
```diff
@@ -80,18 +80,36 @@ pull request is merged. Please see the file [.travis.yml](./.travis.yml)
...
+### `DecimalField` isn't supported
+
+Spanner doesn't support a NUMERIC data type that would allow storing high
+precision decimal values without the possibility of data loss.
+
+## `Variance` and `StdDev` database functions aren't supported
+
+Spanner doesn't have these functions.
+
+### IMO, this section is unclear. What are the practical effects users will care about?
```
timgraham: @odeke-em, please give your input

README.md
```diff
@@ -80,18 +80,36 @@ pull request is merged. Please see the file [.travis.yml](./.travis.yml)
...
+
 Optimistic transactions when running DDL and other statements|Cloud Spanner CANNOT run DDL (CREATE, DROP, ALTER) in a Transaction and CAN ONLY run them in the [DatabaseAdmin RPC]() so any mixes of DDL and other statements will make prior statements get run, DDL run, then following statements run separately|
-107 unexplored Django tests|We need to run every single test in the Django test suite for 100% certainty|Currently in progress to complete the [211 test cases](https://gist.github.com/odeke-em/d3476b6cb7c3485de6aa29956ed50a75#file-all_tests-txt)
+
+### I don't think these issues need to be called out as they're general spanner restrictions that aren't going to be encountered in normal Django usage (if you're specifying db_table, you should know the database's restrictions)
```
timgraham: @odeke-em, please give your input

README.md
```diff
@@ -129,19 +129,26 @@ DJANGO_WORKER_COUNT=1 DJANGO_WORKER_INDEX=0SPANNER_TEST_INSTANCE=django-tests ./
...
 
-Limitation|Comment|Resolution
----|---|---
-Lack of DEFAULT for columns|Cloud Spanner doesn't support using DEFAULT for columns thus the use of default values might have to enforced in your controller logic|
-Lack of FOREIGN KEY constraints|Cloud Spanner doesn't support foreign key constraints thus they have to be defined in code
-Lack of sequential and auto-assigned IDs|Cloud Spanner doesn't autogenerate IDs and this integration instead creates UUID4 to avoid [hotspotting](https://cloud.google.com/spanner/docs/schema-design#uuid_primary_key) so you SHOULD NOT rely on IDs being sorted|We generate UUID4s for each AutoField
-Numeric values are saved as FLOAT64|Cloud Spanner doesn't support the NUMERIC type thus we might have precision losses|Decimal and Numeric are translated to FLOAT64. If Cloud Spanner adds NUMERIC in the future, you might need to migrate your columns
-Optimistic transactions when running DDL and other statements|Cloud Spanner CANNOT run DDL (CREATE, DROP, ALTER) in a Transaction and CAN ONLY run them in the [DatabaseAdmin RPC]() so any mixes of DDL and other statements will make prior statements get run, DDL run, then following statements run separately|
```
timgraham, line 138: Is this still relevant? I thought we only kept the autocommit code now?

README.md
```diff
@@ -129,19 +129,26 @@ DJANGO_WORKER_COUNT=1 DJANGO_WORKER_INDEX=0SPANNER_TEST_INSTANCE=django-tests ./
...
-107 unexplored Django tests|We need to run every single test in the Django test suite for 100% certainty|Currently in progress to complete the [211 test cases](https://gist.github.com/odeke-em/d3476b6cb7c3485de6aa29956ed50a75#file-all_tests-txt)
-Table names CANNOT have spaces within them whether back-ticked or not|Cloud Spanner DOEST NOT support tables with spaces in them for example `Entity ID`|Ensure that your table names don't have spaces within them
```
timgraham, line 140: I don't think table name restrictions need to be called out as they're general Spanner restrictions that aren't going to be encountered in normal Django usage. If someone is customizing a model's db_table, they can read about the database's restrictions in Spanner's docs. I don't think it's the job of django-spanner to duplicate that documentation.

README.md
odeke-em: No longer relevant indeed.

README.md
odeke-em: SGTM!

README.md
```diff
@@ -129,19 +129,56 @@ DJANGO_WORKER_COUNT=1 DJANGO_WORKER_INDEX=0SPANNER_TEST_INSTANCE=django-tests ./
...
-Table names CANNOT have punctuation marks and MUST contain valid UTF-8|Cloud Spanner DOEST NOT support punctuation marks e.g. periods ".", question marks "?" in table names|Ensure that your table names don't have punctuation marks
-Lack of stored procedures and triggers|Cloud Spanner DOES NOT support stored procedures nor triggers so if your code relies on them, you might need to move the logic into your application|Not supported
-Lack of VARIANCE, STDDEV database functions|Cloud Spanned doesn't yet implement these functions|
+### `AutoField` generates random IDs
+
+Spanner doesn't have a way to auto-generate primary key values. Instead,
+django-spanner monkeypatches `AutoField` to generate a random UUID4.
+
+To avoid [hotspotting](https://cloud.google.com/spanner/docs/schema-design#uuid_primary_key),
```
odeke-em, line 137: I think that we need to add that:
>Please don't rely on incrementally increasing numeric ids when sorting.
>You'll need to rewrite your application logic if you rely to sorting by id.

README.md
```diff
@@ -129,19 +129,56 @@ DJANGO_WORKER_COUNT=1 DJANGO_WORKER_INDEX=0SPANNER_TEST_INSTANCE=django-tests ./
 
 ## Limitations
 
-Limitation|Comment|Resolution
----|---|---
-Lack of DEFAULT for columns|Cloud Spanner doesn't support using DEFAULT for columns thus the use of default values might have to enforced in your controller logic|
```
odeke-em, line 134: We definitely need to bring this back, quite important to let users know that they can't rely on defaults.

README.md
```diff
@@ -129,19 +129,56 @@ DJANGO_WORKER_COUNT=1 DJANGO_WORKER_INDEX=0SPANNER_TEST_INSTANCE=django-tests ./
...
-107 unexplored Django tests|We need to run every single test in the Django test suite for 100% certainty|Currently in progress to complete the [211 test cases](https://gist.github.com/odeke-em/d3476b6cb7c3485de6aa29956ed50a75#file-all_tests-txt)
-Table names CANNOT have spaces within them whether back-ticked or not|Cloud Spanner DOEST NOT support tables with spaces in them for example `Entity ID`|Ensure that your table names don't have spaces within them
-Table names CANNOT have punctuation marks and MUST contain valid UTF-8|Cloud Spanner DOEST NOT support punctuation marks e.g. periods ".", question marks "?" in table names|Ensure that your table names don't have punctuation marks
-Lack of stored procedures and triggers|Cloud Spanner DOES NOT support stored procedures nor triggers so if your code relies on them, you might need to move the logic into your application|Not supported
```
odeke-em, line 142: We need to document this too. If users were relying on triggers and stored procedures, they need to also know that Cloud Spanner won't provide them.

timgraham: As we discussed in the design doc, Django doesn't use triggers and stored procedures. I don't think it's the job of django-spanner's README to document how Spanner differs from other relational databases.

README.md
timgraham: Django doesn't set database defaults (except temporarily to set values on new columns; django-spanner uses an UPDATE query to behave equivalently).

README.md
timgraham: I'm not convinced. I think programmers know how random values differ from serial values.

odeke-em: They do but let’s err on the side of caution and explicitness, so that someone reading this guide can just add items to their checklist. I’ve had to call out this fact many times when describing implications of switching backends to customers, and this was to very technical engineers.

README.md
odeke-em: Awesome then, thanks for the update.

README.md
odeke-em: Valid point. However am looking at this from the perspective of being a self contained guide and upfront declaring the effects of switching. This project gets judged in isolation and when folks are switching, they’ll expect that what now doesn’t work, despite Django not having used it, needs to be called out.

timgraham: So we should have a "How Spanner differs from other databases (unrelated to any Django features)" section? This could be very lengthy. I think it makes more sense to defer to Spanner's documentation like https://cloud.google.com/solutions/migrating-oracle-to-cloud-spanner https://cloud.google.com/solutions/migrating-mysql-to-spanner and https://cloud.google.com/spanner/docs/migrating-postgres-spanner but if you want to copy that stuff here, I'll do it.

odeke-em: That would open flood gates, but I really was just requesting for a small general point to be documented. We can leave "how Spanner differs from other databases" for Cloud Spanner writers. I'll resolve this conversation.

odeke-em:
@timgraham one other thing to update is the lack of multi-statement transactions aka there isn't a guarantee that n statements issued in succession will be run in the same transaction. This is a limitation of Spanner but it directly relates to our driver.

timgraham:
Did you see the "Transaction management isn't supported" section? If that's insufficient, I'm unsure how (i.e. how does your statement relate to Django?).

odeke-em:
>Did you see the "Transaction management isn't supported" section? If that's insufficient, I'm unsure how (i.e. how does your statement relate to Django?).

I did, but that's related to Django and django-spanner. I guess what I should have said was, spanner_dbapi (even outside of django-spanner), cannot guarantee multi-statement transactions as that's a limitation of Spanner itself (DDL cannot mix with other statements, INFORMATION_SCHEMA cannot be run in a READ-WRITE transaction), but this reflects too on spanner_dbapi. I can send a PR later on to describe this.

Please rebase from master, then hopefully we can merge this PR in today.

PR #436, opened 2020-04-18
README, examples: add django_spanner as first entry into INSTALLED_APPS

### LINKED ISSUE #433: No mention of INSTALLED_APPS in README
`'django_spanner'` should be added as the first entry in `INSTALLED_APPS` to ensure monkey patching of `AutoField` happens.

odeke-em:
Fixes #433

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,8 +36,17 @@ pip3 install --user .
 ```
 
 ## Using it
-After [installing it](#installing-it), you'll need to edit your Django `settings.py` file
-and particularly the `DATABASES` section to point to an EXISTING database, of the format:
+After [installing it](#installing-it), you'll need to edit your Django `settings.py` file:
+
+* Add `django_spanner` as the very first entry in the `INSTALLED_APPS` setting
+```python
+INSTALLED_APPS = [
+    'django_spanner',
+    ...
+]
+```
+
+* Edit the `DATABASES` setting to point to an EXISTING database
 
 ### Format
 
@@ -66,7 +75,7 @@ DATABASES = {
         'ENGINE': 'django_spanner',
         'PROJECT': 'appdev-soda-spanner-staging', # Or the GCP project-id
         'INSTANCE': 'django-dev1', # Or the Cloud Spanner instance
-        'DATABASE': 'db1', # Or the Cloud Spanner database to use
+        'NAME': 'db1', # Or the Cloud Spanner database to use
     }
 }
 ```
--- a/examples/from-scratch/README.md
+++ b/examples/from-scratch/README.md
@@ -26,7 +26,17 @@ After we have a Cloud Spanner database created, we'll need a few variables:
 * Instance name
 * Database name aka DisplayName
 
-Once in, please edit the file `hc/local_settings.py` and make the section `DATABASES` into the following:
+Once in, please edit the file `hc/local_settings.py`, and:
+
+a) add `django_spanner` as the very first entry to your `INSTALLED_APPS`
+```python
+INSTALLED_APPS = [
+    'django_spanner',  # Must be listed first.
+    ...
+]
+```
+
+b) update `DATABASES` into the following:
 ```python
 DATABASES = {
     'default': {
@@ -56,7 +66,6 @@ DATABASES = {
     }
 }
 ```
-
 
 ### Apply the migrations
 Please run:
--- a/examples/healthchecks/README.md
+++ b/examples/healthchecks/README.md
@@ -61,7 +61,18 @@ After we have a Cloud Spanner database created, we'll need a few variables:
 * Instance name
 * Database name aka DisplayName
 
-Once in, please edit the file `hc/local_settings.py` and make the section `DATABASES` into the following:
+Once in, please edit the file `hc/local_settings.py` to:
+a) Add `django_spanner` as the first entry to `INSTALLED_APPS`
+```python
+INSTALLED_APPS = [
+    'django_spanner',  # Must be listed first.
+    ...
+]
+```
+**The rest of the apps MUST be copied from hc/settings.py**
+
+
+b) Edit `DATABASES` into the following:
 ```python
 DATABASES = {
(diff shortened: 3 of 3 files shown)


README.md
```diff
@@ -36,12 +36,20 @@ pip3 install --user .
 ```
 
 ## Using it
-After [installing it](#installing-it), you'll need to edit your Django `settings.py` file
-and particularly the `DATABASES` section to point to an EXISTING database, of the format:
+After [installing it](#installing-it), you'll need to edit your Django `settings.py` file:
+* Add `django-spanner` as the very first app in the `INSTALLED_APPS` section
```
timgraham, line 40: dash -> underscore
section -> setting

README.md
```diff
@@ -36,12 +36,20 @@ pip3 install --user .
...
+* Edit the `DATABASES` section to point to an EXISTING database, of the format:
```
timgraham, line 41: Perhaps this should be moved after the INSTALLED_APPS example as it looks odd to have a colon and then what follows isn't DATABASES

README.md
```diff
@@ -36,12 +36,20 @@ pip3 install --user .
...
+
+Note: `django_spanner` MUST be added as the first entry into `INSTALLED_APPS`.
```
timgraham, line 43: This seems redundant to two lines earlier.

examples/from-scratch/README.md
```diff
@@ -26,8 +26,15 @@ After we have a Cloud Spanner database created, we'll need a few variables:
 * Instance name
 * Database name aka DisplayName
 
-Once in, please edit the file `hc/local_settings.py` and make the section `DATABASES` into the following:
+Note: `django_spanner` MUST be added as the first entry into `INSTALLED_APPS`.
+
+Once in, please edit the file `hc/local_settings.py` and make the sections `INSTALLED_APPS` and `DATABASES` into the following:
 ```python
+INSTALLED_APPS = [
+    'django_spanner',
```
timgraham, line 34: Rather than a separate note above, perhaps it would be more clear like this:

`'django_spanner',   # Must be listed first.`

examples/healthchecks/README.md
```diff
@@ -82,6 +89,11 @@ and for example here is a filled in database where:
 which when filled out, will look like this
 
 ```python
+INSTALLED_APPS = [
+    'django_spanner',
+    ...
```
timgraham, line 94: Leaving three dots in this section which appears under the words "which when filled out" seems misleading.

timgraham:
(propose to add flaky test back to skip list in https://github.com/googleapis/python-spanner-django/pull/437)

examples/from-scratch/README.md
odeke-em: Ack, thanks!

examples/healthchecks/README.md
odeke-em: Ack, gotcha, thanks!

README.md
odeke-em: Ack, thanks!

README.md
odeke-em: I'll instead add a new fresh sentence because, in two different they will be confusing.

examples/healthchecks/README.md
timgraham: Removing the dots isn't correct. I figured you would put in all the apps that are needed for the tutorial.

examples/healthchecks/README.md
```diff
@@ -82,6 +86,10 @@ and for example here is a filled in database where:
 which when filled out, will look like this
 
 ```python
+INSTALLED_APPS = [
```
timgraham, line 89: When I suggested to remove the dots, I thought you would fill in apps that are needed. `django_spanner` isn't going to be the only app in INSTALLED_APPS. You need to copy all the other apps from the INSTALLED_APPS in hc/settings.py. (Same idea applies to the other examples.)

odeke-em: The problem with that is that all the apps have to be listed, the goal is to just show what needs to be added to the front.

timgraham: I suggest to at least put the three dots back in the first mention of INSTALLED_APPS and remove INTSALLED_APPS from the "here is a filled in database" section (it seems to only distract from the purpose of that).

odeke-em: Ack, thanks!

examples/healthchecks/README.md
odeke-em: Fixed up.

examples/from-scratch/README.md
odeke-em: Moved around.

README.md
```diff
@@ -36,12 +36,23 @@ pip3 install --user .
...
+    ...
+]
+```
+
+* Edit the `DATABASES` setting to point to an EXISTING database
 
 ### Format
 
 ```python
+
```
timgraham, line 54: delete whitespace

examples/healthchecks/README.md
```diff
@@ -61,7 +61,16 @@ After we have a Cloud Spanner database created, we'll need a few variables:
 * Instance name
 * Database name aka DisplayName
 
-Once in, please edit the file `hc/local_settings.py` and make the section `DATABASES` into the following:
+Once in, please edit the file `hc/local_settings.py` to:
+a) Add `django_spanner` as the first entry to `INSTALLED_APPS`
+```python
+INSTALLED_APPS = [
+    'django_spanner',  # Must be listed first.
```
timgraham, line 68: Missing three dots on next line. Perhaps it would also be helpful to mention that the rest of the apps have to be copied from hc/settings.py as local_settings.py doesn't contain the INSTALLED_APPS setting.

odeke-em: Done, thank you!

README.md
odeke-em: Done, thanks!

PR #447, opened 2020-04-29
ci: reduce number of workers for presubmit check

### LINKED ISSUE #444: kokoro builds fail due to 'Instance create requests per minute' of service 'spanner.googleapis.com'
Since https://github.com/googleapis/python-spanner-django/pull/431 was merged (unintentionally), some workers are failing with errors like "Quota exceeded for quota metric 'Instance create requests' and limit 'Instance create requests per minute' of service 'spanner.googleapis.com' for consumer 'project_number:1065521786570'.

We need to decrease the number of workers or get a quota bump.

timgraham:
This reverts commit 7959874a6f09da750533d10f90d94320e082be76.

fixes #444

### DIFF
--- a/.kokoro/build.sh
+++ b/.kokoro/build.sh
@@ -57,20 +57,23 @@ sudo apt-get update
 apt-get install -y libffi-dev libjpeg-dev zlib1g-dev libmemcached-dev
 cd $DJANGO_TESTS_DIR/django && pip3 install -e . && pip3 install -r tests/requirements/py3.txt; cd ../../
 
-
-if [[ $KOKORO_JOB_NAME == *"continuous"* ]]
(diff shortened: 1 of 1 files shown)


timgraham:
Alternatively, I wonder if a solution could be for each worker to start by sleeping for something like (X *  worker #) seconds so as not to exceed the "Instance create requests per minute" cap (whatever that limit is).

odeke-em:
We already do that
https://github.com/googleapis/python-spanner-django/blob/3264f75e7da336937fd4d4c3656b8d3ade9f58ea/parallelize_tests.go#L174-L180

I think that continuous still runs, so earlier this morning I kindly asked @busunkim96 to help us entirely delete it from the Kokoro UI.

Issue #449, opened 2020-05-12
package publishing: use pypi


odeke-em:
We'll need to publish django-spanner on pypi.

Kindly paging @busunkim96 @skuruppu @vmanghnani

skuruppu:
@larkee was looking into setting up ReleasePlease yesterday. Let us know if you need any help.

busunkim96:
I added conventional commit bot to the repo, so you will see it start to see a status check on PRs (non-blocking). The bot uses commits to propose new releases. You will need a commit of type `fix` or `feat` for the bot to create a new PR.  

https://www.conventionalcommits.org/en/v1.0.0/

Besides that I think all the machinery is already in this repo to do release builds.

skuruppu:
Oh cool, didn't know. Thanks heaps @busunkim96.

skuruppu:
Downgrading the priority to appease the bots. We can't publish a release yet as the alpha release is still awaiting the internal approvals process.

vi3k6i5:
With beta releases now we are following `ReleasePlease` process and new packages are getting released on pypi.

PR #454, opened 2020-05-19
fix: update pypi package name

skuruppu:
Towards #449

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -4,6 +4,7 @@
 # license that can be found in the LICENSE file or at
 # https://developers.google.com/open-source/licenses/bsd
 
+import io
 import os
 
 from setuptools import find_packages, setup
--- a/setup.py
+++ b/setup.py
@@ -35,9 +35,6 @@ with io.open(readme_filename, encoding="utf-8") as readme_file:
 
 setup(
         name=name,
-        # Duplicate version here rather than using
-        # __import__('django_spanner').__version__ because that file imports
-        # django and google.cloud which may not be installed.
         version=version,
         description=description,
         long_description=readme,
--- a/setup.py
+++ b/setup.py
@@ -6,7 +6,25 @@
 
 from setuptools import find_packages, setup
 
-version = '2.2a0'
+
+# Package metadata.
+
+name = "django-google-spanner"
+description = "Bridge to enable using Django with Spanner."
+version = "2.2a0"
+# Should be one of:
+# 'Development Status :: 3 - Alpha'
+# 'Development Status :: 4 - Beta'
+# 'Development Status :: 5 - Production/Stable'
+release_status = "Development Status :: 3 - Alpha"
+dependencies = [
+    'sqlparse >= 0.3.0',
+    'google-cloud-spanner >= 1.8.0',
+]
+extras = {}
+
+
+# Setup boilerplate below this line.
 
 package_root = os.path.abspath(os.path.dirname(__file__))
 
@@ -14,27 +32,23 @@ readme_filename = os.path.join(package_root, "README.rst")
 with io.open(readme_filename, encoding="utf-8") as readme_file:
     readme = readme_file.read()
 
-install_requires = [
-    'sqlparse >= 0.3.0',
-    'google-cloud-spanner >= 1.8.0',
-]
 
 setup(
-        name='django-google-spanner',
+        name=name,
         # Duplicate version here rather than using
         # __import__('django_spanner').__version__ because that file imports
         # django and google.cloud which may not be installed.
         version=version,
+        description=description,
+        long_description=readme,
         author='Google LLC',
         author_email='cloud-spanner-developers@googlegroups.com',
-        description=('Bridge to enable using Django with Spanner.'),
-        long_description=readme,
         license='BSD',
         packages=find_packages(exclude=['tests']),
-        install_requires=install_requires,
+        install_requires=dependencies,
         url="https://github.com/googleapis/python-spanner-django",
         classifiers=[
-            'Development Status :: 4 - Beta',
+            release_status,
             'Environment :: Web Environment',
             'Intended Audience :: Developers',
             'License :: OSI Approved :: BSD',
@@ -47,5 +61,6 @@ setup(
             'Framework :: Django',
             'Framework :: Django :: 2.2',
         ],
+        extras_require=extras,
         python_requires=">=2.7,!=3.0.*,!=3.1.*,!=3.2.*,!=3.3.*",
 )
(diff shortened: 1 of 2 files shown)


odeke-em:
I shall later on send docs updates to indicate the new name.

setup.py
```diff
@@ -10,12 +10,11 @@
 
 install_requires = [
     'sqlparse >= 0.3.0',
-    'google-cloud >= 0.34.0',
```
skuruppu, line 13: @odeke-em, @busunkim96 asked me to remove this dependency. Hope that's ok with you.

busunkim96: For context, this is an old package (last published July 2018) https://pypi.org/project/google-cloud/. The cloud libraries used to all be published under this single package but were split out.

odeke-em:
What’s the rationale for removing it? It is necessary for folks who’ll be
using this package to install once and not have to manually install
google-cloud as well. I could be wrong though, and if there is another way
to do it then sure.

setup.py
```diff
@@ -10,12 +10,11 @@
...
     'google-cloud-spanner >= 1.8.0',
 ]
 
 setup(
-        name='django-spanner',
+        name='django-google-spanner',
         # Duplicate version here rather than using
```
busunkim96, line 18: Please single-source the version. You can treat the version in this file as the source of truth and get the version using `pkg_resources.get_distribution()` in the codebase.

[example in python-spanner](https://github.com/googleapis/python-spanner/blob/55a21d97d0c863cbbbb2d973b6faa4aeba8e38bb/google/cloud/spanner_v1/__init__.py#L19)

setup.py
odeke-em: Gotcha, thanks! In that case, can we replace it with the appropriate cloud-Spanner import?

odeke-em: Oh never mind, we already Gabe the google-cloud-spanner dependency. Resolving this conversation as I’ve answered my own question.

setup.py
skuruppu: All done.

skuruppu:
> I think this setup.py would benefit from having a few more fields:
> 
> * `url`
> * `python_requires` (pip won't install the package if the Python version is incompatible)
> * `long_description`
> 
> See [google-cloud-spanner's setup.py](https://github.com/googleapis/python-spanner/blob/master/setup.py#L59-L89) for comparison

I made setup.py look similar to https://github.com/googleapis/python-spanner/blob/master/setup.py.

setup.py
```diff
@@ -6,28 +6,46 @@
...
 from setuptools import find_packages, setup
 
-version = '2.2a0'
 
-install_requires = [
+# Package metadata.
+
+name = "django-google-spanner"
+description = "Bridge to enable using Django with Spanner."
+version = "2.2a0"
```
skuruppu, line 14: I assume release please will replace this version since this doesn't look like the version number it's proposing in #453

odeke-em:
If possible I’d highly encourage we stick to the 2.2a0 format that
currently exists because it contains what version of Django (2.2) is
supported but also what version of the release is in (a0)

setup.py
```diff
@@ -39,4 +58,6 @@
             'Framework :: Django',
             'Framework :: Django :: 2.2',
         ],
+        extras_require=extras,
+        python_requires=">=2.7,!=3.0.*,!=3.1.*,!=3.2.*,!=3.3.*",
```
busunkim96, line 62: @odeke-em Is this library compatible with 2.7?

odeke-em:
Nope, it is meant for Python 3.5 and above.

busunkim96:
re: versions

I would prefer semver for consistency with other packages. It also makes it easier for folks to pin on this library (e.g., `google-django-spanner>=0.1.0,<1.0.0` If you all decide on some other version schema please explain it clearly in the README. The release bot won't be able to figure out what the correct 'next' release is, so you will have to manually propose releases.

1. Open a PR. In the body of the commit (or the squash merge commit), add 'Release-As: x.x."

```
feat: release 2.2a0

Release-As: 2.2.a0
```

2. Wait for Release-Please to open a new PR.

skuruppu:
@busunkim96, for reducing the maintenance burden on my team, I don't want to deviate from the versioning done by other python packages in googleapis. We want to automate as much of the chores as possible.

@odeke-em, this is how other [googeapis django](https://www.google.com/search?q=django+google+pypi&rlz=1CAOFPC_enAU814&oq=django+google+pypi&aqs=chrome..69i57j0l2j69i60l3j69i64l2.6894j0j7&sourceid=chrome&ie=UTF-8) releases are versioned (if you click through the search links).

setup.py
skuruppu: Updated to >=3.5.

setup.py
timgraham: That wouldn't be ideal. The README explains the typical versioning of database backends: "Use the version of django-spanner that corresponds to your version of Django. For example, django-spanner 2.2.x works with Django 2.2.y. (This is the only supported version at this time.)"

skuruppu: @timgraham thanks for that info. But see my [comment](https://github.com/googleapis/python-spanner-django/pull/454#issuecomment-631191745). There are lots of existing django packages that are published by Google and they all consistently use the versioning scheme used by release-please.

timgraham: Yes, sem-var versioning is typical for other Django libraries that often support multiple versions of Django. Database backends are different. The best practice is to create separate release branches for each version of Django that's supported.

skuruppu: I spoke to @larkee and they explained that the manual releasing won't be so painful.

> re: versions
> 
> I would prefer semver for consistency with other packages. It also makes it easier for folks to pin on this library (e.g., `google-django-spanner>=0.1.0,<1.0.0` 

@timgraham do you have any thoughts on the issue of pinning releases when you have non-semver versioning?

timgraham: If using Django 2.2.x, use `django-google-spanner>=2.2,<3.0`. I think `django-google-spanner==2.2.*` would be equivalent.

skuruppu: In that case, @timgraham or @odeke-em could you update the README to explain why the versioning is the way it is? Please include links to the Django README as a reference.

timgraham: What I quoted is from the README in this repo.

setup.py
```diff
@@ -6,28 +6,46 @@
...
     'sqlparse >= 0.3.0',
-    'google-cloud >= 0.34.0',
     'google-cloud-spanner >= 1.8.0',
 ]
+extras = {}
+
+
+# Setup boilerplate below this line.
+
+package_root = os.path.abspath(os.path.dirname(__file__))
```
larkee, line 29: You need to `import os`. The presubmit tests are all failing because `os` is not defined

PR #456, opened 2020-05-21
README: document how we version releases

odeke-em:
Document why the release in setup.py looks like
```python
version = '2.2a0'
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,6 +20,7 @@ ORM plugin for using Cloud Spanner as a database for Django.
 - [How it works](#how-it-works)
     - [Overall design](#overall-design)
     - [Internals](#internals)
+- [Versioning](#versioning)
 
 
 ## Installing it
@@ -199,5 +200,15 @@ NULL ...
 
 ### Internals
 ![](./assets/internals.png)
+
+## Versioning
+The release version of django-spanner in [setup.py](./setup.py) must follow this form
+```python
+version = '<DJANGO_VERSION><RELEASE_MODE><RELEASE_VERSION>'
+```
+for example, if this package supports Django version 2.2.X, and we are in alpha release mode, while this application's release is version 0, our version will be
+```python
+version = '2.2a0'
+```
 
 # 🚨THIS CODE IS STILL UNDER DEVELOPMENT🚨


README.md
```diff
@@ -200,4 +201,14 @@ NULL ...
 ### Internals
 ![](./assets/internals.png)
 
+## Release versioning
+The release version of django-spanner in [setup.py](./setup.py) SHOULD follow this form
```
skuruppu, line 205:
```suggestion
The release version of django-spanner in [setup.py](./setup.py) follows this format:
```

README.md
```diff
@@ -200,4 +201,14 @@ NULL ...
...
+```python
+version = '<DJANGO_VERSION><RELEASE_MODE><RELEASE_VERSION>'
+```
+for example, if this package supports Django version 2.2.X, and we are in alpha release mode, while this application's release is version 0, our version will be
```
skuruppu, line 209:
```suggestion
For example, if this package supports Django version 2.2.X, and we are in alpha release mode, while this application's release is version 0, our version will be:
```

README.md
odeke-em: I'd encourage we include the `SHOULD` as it is the way that django-backends must be versioned.

README.md
odeke-em: Ack, thank you!

README.md
skuruppu: Ok in that case, "must" but there's no need to capitalize it.

skuruppu, line 205: I would still like you to add a reference to some documentation, either a PEP or something more Django-specific that expects us to follow this convention.

I looked through [PEP 440](https://www.python.org/dev/peps/pep-0440) and I didn't find anything relevant but you may have a different reference.

README.md
```diff
@@ -200,4 +201,14 @@ NULL ...
 ### Internals
 ![](./assets/internals.png)
 
+## Release versioning
```
skuruppu, line 204:
```suggestion
## Versioning
```

See this [client library README](https://github.com/googleapis/nodejs-spanner/blob/831846cd86e91fdb343ac2f64a6b257877e0e157/.kokoro/presubmit/node10/system-test-against-emulator.cfg) as an example of how to explain versioning in the README.

README.md
odeke-em: Could we perhaps dig up the reference after merging this PR and releasing? I got the information and guidance from Tim whom I'll consult to dig up the reference.

README.md
odeke-em: Have we yet decided if we are promoting this to alpha, beta etc? I thought for now we are documenting the basic version just to get a release.

README.md
skuruppu: The release is still pending approval so if we can complete the README in the meantime, that would be great.

README.md
skuruppu: The version documentation is independent of releases. It's just a section in the docs to show what release scheme we're using. So it'll be whatever defines the <django version><pre-release><version> format.

In any case, we'll always start with alpha. I don't think we want to publish something before alpha to pypi.

README.md
```diff
@@ -200,4 +201,14 @@ NULL ...
...
 ![](./assets/internals.png)
 
+## Versioning
+The release version of django-spanner in [setup.py](./setup.py) must follow this form
+```python
+version = '<DJANGO_VERSION><RELEASE_MODE><RELEASE_VERSION>'
+```
+for example, if this package supports Django version 2.2.X, and we are in alpha release mode, while this application's release is version 0, our version will be
+```python
+version = '2.2a0'
```
timgraham, line 211: a0 means "pre-alpha" (never released). For the first alpha release, it should be a1.

README.md
```diff
@@ -200,4 +201,14 @@ NULL ...
 ### Internals
 ![](./assets/internals.png)
 
+## Versioning
```
timgraham, line 204: This section seems better in CONTRIBUTING.md. The README is targeted at users who don't care about this. (It's already covered above in the statement "Use the version of django-spanner that corresponds to your version of Django.
For example, django-spanner 2.2.x works with Django 2.2.y. (This is the only supported version at this time.)")

PR #462, opened 2020-08-21
feat: refactor connect() function, cover it with unit tests

### LINKED ISSUE #460: connect() surface issues
Looking at `connect()` function, there are several things which are raising questions.

**project arg:**
 `project` is set as an optional arg: https://github.com/googleapis/python-spanner-django/blob/11222db2f82fd50ca87010321ded0b39021eb884/spanner_dbapi/__init__.py#L34 But few lines lower it is checked for existence (thus, it's required, not optional): https://github.com/googleapis/python-spanner-django/blob/11222db2f82fd50ca87010321ded0b39021eb884/spanner_dbapi/__init__.py#L51-L52 It seems to be pointless, as this arg is only passed to Spanner client, and the original Spanner client can handle an omitted project: it'll use `google.auth.default()` function to designate the project id from the environment. With this, I'd propose to drop this check and use `project` arg in the way of the original Spanner client.

**instance and database args:**
They are optional by definition, but in fact both are required. Probably, they should become required in definition (the function signature will be `connect(instance, database, project=None, credentials_uri=None, user_agent=None):`), so these checks could be dropped:
https://github.com/googleapis/python-spanner-django/blob/11222db2f82fd50ca87010321ded0b39021eb884/spanner_dbapi/__init__.py#L53-L56

**credentials_uri arg:**
This arg looks divergent from how the original Spanner client is dealing with credentials:
https://github.com/googleapis/python-spanner/blob/b6b0348f5870f55c88b3cae44e229325fd755b26/google/cloud/spanner_v1/client.py#L123-L129

I'd propose to change this arg to use `google.auth.credentials.Credentials` class as it's done in the original Spanner client.

IlyaFaer:
Run formatters.
Fit doclines, variable names and function args to the original Spanner client fashion.
Cover the function with unit tests.

Closes #460


spanner_dbapi/__init__.py
```diff
@@ -4,83 +4,122 @@
...
+    Create a connection to Cloud Spanner database.
 
-    Args:
-        project: The id of a project that already exists.
-        instance: The id of an instance that already exists.
-        database: The name of a database that already exists.
-        credentials_uri: An optional string specifying where to retrieve the service
-                         account JSON for the credentials to connect to Cloud Spanner.
+    :type instance_id: :class:`str`
+    :param instance_id: ID of the instance to connect to.
```
c24t, line 56: Nit: why change the docstring style? It looks like this repo uses google-style docstrings, and we can use [napoleon](https://www.sphinx-doc.org/en/master/usage/extensions/napoleon.html) to handle docs generation.

spanner_dbapi/__init__.py
```diff
@@ -4,83 +4,122 @@
 # license that can be found in the LICENSE file or at
 # https://developers.google.com/open-source/licenses/bsd
 
-from google.cloud import spanner_v1 as spanner
+"""Connection-based DB API for Cloud Spanner."""
+
+from google.cloud import spanner_v1
```
c24t, line 9: 👍

spanner_dbapi/__init__.py
```diff
@@ -4,83 +4,122 @@
...
+        credentials=credentials,
+        client_info=google_client_info(user_agent),
+    )
 
-    client_instance = client.instance(instance)
-    if not client_instance.exists():
-        raise ProgrammingError("instance '%s' does not exist." % instance)
+    instance = client.instance(instance_id)
+    if not instance.exists():
+        raise ProgrammingError("instance '%s' does not exist." % instance_id)
```
c24t, line 86: Looks to me like a `ValueError` since it could be the caller's fault if they pass in the wrong credentials or instance ID.

I know you didn't make this change in the first place, but may as well fix this since you're refactoring.

tests/spanner_dbapi/test_connect.py
```diff
@@ -0,0 +1,116 @@
...
+
+    class _CredentialsWithScopes(
+        google.auth.credentials.Credentials, google.auth.credentials.Scoped
+    ):
+        pass
+
+    return mock.Mock(spec=_CredentialsWithScopes)
+
+
+class Testconnect(unittest.TestCase):
```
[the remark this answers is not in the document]
c24t: Not `Test_connect`?

tests/spanner_dbapi/test_connect.py
```diff
@@ -0,0 +1,115 @@
...
+        return connect(*args, **kw)
+
+    def test_connect(self):
+        from google.api_core.gapic_v1.client_info import ClientInfo
+        from spanner_dbapi.connection import Connection
+
+        PROJECT = "test-project"
+        USER_AGENT = "user-agent"
+        CREDENTIALS = _make_credentials()
+        CLIENT_INFO = ClientInfo(user_agent=USER_AGENT)
```
c24t, line 37: Nit: why the caps if these are just regular non-global vars?

tests/spanner_dbapi/test_connect.py
```diff
@@ -0,0 +1,115 @@
...
+        google.auth.credentials.Credentials, google.auth.credentials.Scoped
+    ):
+        pass
+
+    return mock.Mock(spec=_CredentialsWithScopes)
+
+
+class Testconnect(unittest.TestCase):
+    def _callFUT(self, *args, **kw):
+        from spanner_dbapi import connect
```
c24t, line 26: Why do you need to bury the import here instead of just doing `import spanner_dbapi` above?

tests/spanner_dbapi/test_connect.py
```diff
@@ -0,0 +1,115 @@
...
+            ):
+                connection = self._callFUT("test-instance", DATABASE)
+
+                database_mock.assert_called_once_with(DATABASE, pool=mock.ANY)
+
+        self.assertIsInstance(connection, Connection)
+
+
+if __name__ == "__main__":
+    unittest.main()
```
c24t, line 115: I don't see this in the other tests, I don't think the test modules need to be individually executable.

c24t:
> I don't see `isort` mentions in the original Spanner client. Maybe we should redact `tox` to run `black` instead of `isort` (in this PR probably)?

Another option is to change isort's config to appease black. 

```
[settings]
include_trailing_comma=True
force_grid_wrap=0
use_parentheses=True
line_length=79
multi_line_output=3
skip=target
skip_glob=...
known_first_party=...
known_third_party=...
```

We had to do this in another project here: https://github.com/open-telemetry/opentelemetry-python/pull/104#discussion_r321004334.

spanner_dbapi/__init__.py
IlyaFaer: Such a style is used in the original Spanner repo ([example](https://github.com/googleapis/python-spanner/blob/45a15382bc1e62dedc944f6484c15ba929338670/google/cloud/spanner_v1/client.py#L83-L90)). Just mimicking, as we probably will have to change to the original client style when moving into it. I am myself okay with both styles

spanner_dbapi/__init__.py
IlyaFaer: Done

tests/spanner_dbapi/test_connect.py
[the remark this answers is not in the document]
IlyaFaer: Looks like `_` is used only because the function name itself starts with a `_` symbol: `_make_list_value_pbs`. Still, I don't think it's a strict naming rule, and I like `Test_connect` more. Let's rename

tests/spanner_dbapi/test_connect.py
IlyaFaer: I don't like it myself, and when I first saw it, it made me confused. But that is what they do in Google Cloud ([example](https://github.com/googleapis/python-spanner/blob/45a15382bc1e62dedc944f6484c15ba929338670/tests/unit/test_database.py#L208-L211)). I think it's a way to avoid using instances patched with mocks by another tests. If it's okay to use a single import at the top of the file, I'm definitely for it.

tests/spanner_dbapi/test_connect.py
IlyaFaer: It's what they do in other clients ([example](https://github.com/googleapis/python-spanner/blob/45a15382bc1e62dedc944f6484c15ba929338670/tests/unit/test_database.py#L213-L214)). I can't say why it's done like this, but I'm okay with both variants.

tests/spanner_dbapi/test_connect.py
IlyaFaer: Erased

IlyaFaer (commented):
@c24t, looks like comments are mostly about things that I'm just mimicking from other Cloud repos. It's not my preferences, so I'm ready to do however you (as a googler) decide

spanner_dbapi/__init__.py
c24t: Ah ok. Not a strong preference, but I think it makes sense to keep this repo all one style for now, even when it conflicts with python-spanner.

tests/spanner_dbapi/test_connect.py
c24t: That's interesting, I see the python-spanner tests bury a lot of imports, including things (like the proto in your example) that you wouldn't really expect others to mock or to have import-time effects.

Maybe there's a good reason to do this, but the test runs fine without it. I think in general it's better to have code that's easy to understand, and if we find out that this doesn't work we can copy the pattern from python-spanner.

FWIW `_callFUT` goes all the way back to the first public release of python-spanner (https://github.com/googleapis/python-spanner/commit/fb9733b), I don't see any justification for using it in PRs or git history.

IlyaFaer: SG, fixed all of it

django_spanner/base.py
```diff
@@ -104,19 +106,19 @@ class DatabaseWrapper(BaseDatabaseWrapper):
...
     def get_connection_params(self):
         return {
-            'project': self.settings_dict['PROJECT'],
-            'instance': self.settings_dict['INSTANCE'],
-            'database': self.settings_dict['NAME'],
-            'user_agent': 'django_spanner/0.0.1',
-            **self.settings_dict['OPTIONS'],
+            "project": self.settings_dict["PROJECT"],
+            "instance_id": self.settings_dict["INSTANCE"],
+            "database_id": self.settings_dict["NAME"],
```
IlyaFaer, line 119: Changed arg names here, other changes in the file are just about formatting

PR #463, opened 2020-08-24
feat: cursor must detect if the parent connection is closed

### LINKED ISSUE #461: Connection.close() doesn't close cursors
According to PEP-249, on connection closing all the related cursors must become no-op as well. The current implementation of closes and checks has an issue at this point.

`Connection.close()` only sets the connection attribute to `True`:
https://github.com/googleapis/python-spanner-django/blob/11222db2f82fd50ca87010321ded0b39021eb884/spanner_dbapi/connection.py#L112-L115

While the `Cursor` checking method is only looking for the `Cursor._closed` attribute:
https://github.com/googleapis/python-spanner-django/blob/11222db2f82fd50ca87010321ded0b39021eb884/spanner_dbapi/cursor.py#L219-L224

That means cursor don't know if the related connection is already closed. Technically, it doesn't look like a big problem, as `Cursor.execute()` delegates to `Connection`, and the connection will raise an exception `already closed`. But there are several commands, which will be runned on a closed `Cursor` before the exception raised, for example, `classify_stmt(sql)`.

To avoid that it'll be good to add into `Cursor._raise_if_already_closed()` a line like:
```python
if self._connection.is_closed:
    raise ...
```
Checking a bool is much more faster than regexping SQL code and passing it into a connection method.

Plus to this, I'd make `is_closed` attribute public, so that users could check if a cursor/connection is closed.

IlyaFaer:
Closes #461


IlyaFaer:
This PR is based on changes from [connect() refactoring changes](https://github.com/googleapis/python-spanner-django/pull/462), so it should be merged after "connect() refactoring" PR

tests/spanner_dbapi/test_cursor.py
```diff
@@ -0,0 +1,46 @@
...
+        ):
+            with mock.patch(
+                "google.cloud.spanner_v1.database.Database.exists", return_value=True
+            ):
+                connection = connect("test-instance", "test-database")
+
+        cursor = connection.cursor()
+        cursor.close()
+
+        self.assertTrue(cursor.is_closed)
```
c24t, line 28: Here and below, might be worth adding an `assertFalse` check for the same thing before calling `close`.

c24t:
Seeing more of these in CI:

`panic: rpc error: code = ResourceExhausted desc = Project 1065521786570 cannot add 1 nodes in region us-west1.`

Hopefully to be fixed in #471.

spanner_dbapi/cursor.py
```diff
@@ -216,32 +220,48 @@ def description(self):
...
 
-    def _raise_if_already_closed(self):
+    @property
+    def is_closed(self):
+        """The cursor close indicator.
+
+        Returns:
+            bool:
+                True if this cursor or it's parent connection
+                is closed, False otherwise.
```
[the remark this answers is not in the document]
IlyaFaer: Oh, wait, I've written it in Napoleon style - that's not correct. Changed to the original Spanner client style.

tests/spanner_dbapi/test_cursor.py
IlyaFaer: Done

IlyaFaer (commented):
I've added few more docstrings/renaming changes

PR #464, opened 2020-08-24
feat: avoid adding a dummy WHERE clause into UPDATE/DELETE queries

IlyaFaer:
[Adding a WHERE clause](https://github.com/googleapis/python-spanner-django/blob/11222db2f82fd50ca87010321ded0b39021eb884/spanner_dbapi/parse_utils.py#L400) into user's query doesn't seem to be a good practise in the first place. More than that Spanner requires WHERE clause presence in UPDATE and DELETE queries (see [the docs](https://cloud.google.com/spanner/docs/dml-syntax#where_clause)). Adding a dummy WHERE clause silently overrides the original Spanner requirement, and actually leads to what the requirement prevents: accidental updating/deleting every row.

It doesn't look correct, let's raise an error in case of no WHERE clause detected - it'll warn user that the thing he's trying to do can be dangerous. If he's okay with it, he can add `WHERE true` by his own hands.


IlyaFaer:
This PR is based on changes from [the previous PR](https://github.com/googleapis/python-spanner-django/pull/463), and should be merged after merging the previous one.

spanner_dbapi/cursor.py
```diff
@@ -77,39 +80,38 @@ def execute(self, sql, args=None):
...
-        self._connection.in_transaction(
-            self.__do_execute_update,
-            sql, params,
-        )
+    def _handle_update(self, sql, params):
+        ensure_where_clause(sql)
+        self._connection.in_transaction(self.__do_execute_update, sql, params)
 
     def __do_execute_update(self, transaction, sql, params, param_types=None):
-        sql = ensure_where_clause(sql)
```
IlyaFaer, line 102: I also realized that it's not very okay to do the check here, as `__do_execute_update()` will be sent to `Database.run_in_transaction()`, and then to `Session.run_in_transaction()` as `func` arg. This last method will start a transaction (if needed) and only then it'll run `__do_execute_update()` (and the check). We better run the check before starting anything. The check call moved to line 99

PR #465, opened 2020-08-24
fix: `BSD` license replaced with `Apache 2.0`

mf2199:
Replacing `BSD` license headers with Google-standard `Apache 2.0`

Fixes: #467

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,6 @@ build/
 MANIFEST
 dist/
 django_tests
+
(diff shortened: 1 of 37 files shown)


c24t:
We may well want this change in the future, but for now we're using BSD because that's what django itself uses. Closing this, but we can revisit if we need to change the license (e.g. to match [python spanner](https://github.com/googleapis/python-spanner)) down the line.

Issue #474, opened 2020-08-28
Add nox target for django tests


c24t:
Right now we run django's built in tests in CI using `python runtests.py`, but it looks like pip-installing django doesn't give us an executable `runtests.py` without us changing `PYTHONPATH`.

Kokoro seems to get around this by installing django from source, from @timgraham's fork: https://github.com/googleapis/python-spanner-django/blob/bfde2214ded7f7205286f62d5a5feac8687f0139/.kokoro/build.sh#L60

Running the [existing tox target](https://github.com/googleapis/python-spanner-django/blob/11222db2f82fd50ca87010321ded0b39021eb884/tox.ini#L6-L15) (from 11222db) doesn't appear to work:

```
.../python-spanner-django/.tox/py37-django22/bin/python: can't open file 'runtests.py': [Errno 2] No such file or directory
ERROR: InvocationError for command .../python-spanner-django/.tox/py37-django22/bin/python runtests.py (exited with code 2)
ERROR:   py37-django22: commands failed
```

Since @mf2199 replaced tox with nox in #468, we need a nox target to run the built-in django tests. This target shouldn't depend on kokoro scripts (like [`django_test_suite.sh` does now](https://github.com/googleapis/python-spanner-django/blob/bfde2214ded7f7205286f62d5a5feac8687f0139/django_test_suite.sh#L50)), and the existing scripts should be changed to remove dependencies on the fork if possible.

See also #471.

vi3k6i5:
Added in https://github.com/googleapis/python-spanner-django/pull/610

PR #466, opened 2020-08-24
feat: [WIP] Migration to `nox` testing automation framework

mf2199:
Since `nox` is the primary testing automation tool used by Google API, we believe that its earliest integration with the Spanner-Django project is necessary to ensure its continuous compliance with the coding standards from the start. This is also a step towards future merging of Django-related code into the main Spanner repository and may save considerable amount of time and effort at later stages of development.

Major change list:

1. Added new file `noxfile.py` [largely copied from Spanner main repo];
2. Added missing folders:
- `docs`
- `google`
- `google/cloud`
- `docs/_static`

3. Folders renamed/moved:
- `django_spanner` --> `google/cloud/spanner_django`;
- `spanner_dbapi` --> `google/cloud/spanner_dbapi`;
- `tests/spanner_dbapi` --> `tests/unit`;
- `examples` --> `samples`;

4. Added relevant `__init__.py` and `conf.py` files;
5. Added docs styling definitions, `custom.css` [copied from main Spanner repo];
6. Resolved missing dependencies in `setup.py`;
7. `README.md` converted to `README.rst` [fixing some formatter `check` errors];
8. Added `docs/index.rst` file to comply with Sphinx requirements (the file is blank and to be updated as necessary);
9. Updated `.gitignore`.

**From the results of the initial testing, the code has numerous assertion errors and coverage gaps, see the test log below, which have little to do with the test setup as such. Therefore, addressing those will be a subject of subsequent PRs, if and when the matters of the current proposition are reviewed and agreed upon.**

Towards #474.

PR #480, opened 2020-08-31
feat: Stage 6 of `nox` implementation - enabling system tests

mf2199:
As suggested in #466 , this represents the last item of the recommended list aimed to implement nox testing automation in multiple steps.

Change list:
1. Added a copy of a common `test_utils` package directory;
2. Added `system` test session to Nox configuration;
3. New `tests/system/test_system.py` template file;
3. Some updates to to `.gitignore` file.

_A dummy test method inside the template file is necessary to avoid an `exit code 5` error, that is when no tests get collected._

Session test log:
```
nox > Running session system
nox > Creating virtual environment (virtualenv) using python.exe in .nox\system
nox > pip install mock pytest
nox > pip install -e .
nox > pip install -e test_utils
nox > py.test --quiet tests\system
.                                                                                                                                                                                                                                [100%]
1 passed in 0.03s
nox > Session system was successful.
```

Towards #474.


noxfile.py
```diff
@@ -87,6 +87,43 @@ def unit(session):
...
+    # Install dependencies, e.g.:
+    # session.install("google-cloud-iam")
+    # Use pre-release gRPC for system tests.
+    # session.install("--pre", "grpcio")
+
+    # Install all test dependencies, then install this package into the
+    # virtualenv's dist-packages.
+    session.install("mock", "pytest")
+    session.install("-e", ".")
+    session.install("-e", "test_utils")
```
busunkim96, line 116: If any of the system tests need the contents of test_utils, you can install it through pip.

It might not be necessary though (there's not too much in it (https://github.com/googleapis/python-test-utils/tree/master/test_utils) at this point.

```
pip install google-cloud-testutils
```

mf2199: @busunkim96 👍 Agreed. This way used to be prevalent across other products and still exists in [Spanner](https://github.com/googleapis/python-spanner/blob/a79786ec3620da21aa3ce1c8bc820dab5983531d/noxfile.py#L126), [Bigtable](https://github.com/googleapis/python-bigtable/blob/8f764343e01d50ad880363f5a4e5630122cbdb25/noxfile.py#L133), and possibly some others. Would it be a good idea to update those as well?

noxfile.py
```diff
@@ -87,6 +87,42 @@ def unit(session):
...
+
+    # Sanity check: only run tests if found.
+    if not system_test_exists and not system_test_folder_exists:
+        session.skip("System tests were not found")
+
+    # Install dependencies, e.g.:
+    # session.install("google-cloud-iam")
+    # Use pre-release gRPC for system tests.
+    # session.install("--pre", "grpcio")
+
```
c24t, line 111:
```suggestion
```

c24t (approved):

PR #481, opened 2020-09-03
feat: reworked `Connection` class

paul1319:
Work in progress


IlyaFaer:
@paul1319, we need to speed up with this PR, as it contains a lot of changes that "should be"

PR #482, opened 2020-09-07
chore: Code refactoring to follow common Google API scheme

mf2199:
Change list:

1. `django_spanner` folder renamed to `spanner_django` and moved inside `google/cloud/`;
2. `spanner_dbapi` folder moved inside `google/cloud/`;
3. Invalidated references updated to reflect the above changes;
4. Relevant updates to `noxfile.py`.

**While working on the build errors, this PR is kept as a draft.**


IlyaFaer (commented):
Looks okay

c24t:
This PR changes the distribution name from `django-google-spanner` to `google-cloud-spanner-django` to match the other googleapis projects. Thankfully this project hasn't been released on PyPI yet, so we don't have to worry about deprecating the old name.

It also changes the package namespace. Now instead of a single top-level `django_spanner` we've got two new namespaces under the shared `google.cloud`: `google.cloud.spanner_django` and `google.cloud.spanner_dbapi`. This is (1) and (2) in the description, I just wanted to make it explicit that we're using namespace packages here and this isn't just a directory move.

Both great changes, and both following the conventions of the other python repos.

c24t:
After talking to @skuruppu it sounds like we should keep the `django-google-spanner` distribution name:
- The `google-cloud-` name usually denotes a GCP service client, and this isn't exactly that.
- Most django apps are named `django-`. This is only convention, but a pretty strong one.
- Other google-owned django apps (like https://pypi.org/project/django-cloud-deploy/) follow this convention too.
- And so do other [third party DB django apps](https://docs.djangoproject.com/en/3.1/ref/databases/#using-a-3rd-party-database-backend).

IlyaFaer:
@mf2199, this PR looks not very actual - most part is already merged. Should we close it?

PR #483, opened 2020-09-09
refactor: update get_param_types() function

IlyaFaer:


IlyaFaer:
@c24t, I've got an error here:
```
KeyError: <class 'django.utils.safestring.SafeText'>
```
Seems like this class was added in Django 3.x. Is it okay to add this type into our types map? I mean, it's a Django specific type, should it appear in DB API (which is not Django-specific)?

PR #488, opened 2020-09-14
refactor: update classify_stmt() helper and cover it with unit tests

IlyaFaer:


c24t (commented):
The new lines in the test look fine, but why the name change? Not all these statements are queries right?

What's the motivation for this change?

IlyaFaer:
@c24t, "statements" sounds too vague for me. For example, "if-else" are usually called statements too. "Query" in this case looks closer to databases.

IlyaFaer:
@c24t

> Not all these statements are queries right?

Not sure what do you mean. I know, DDL are usually called statements, but I suppose it's mostly about operators. I mean `CREATE TABLE` is statement, but `CREATE TABLE "table"` looks like a query to me, especially if it is a big script with several commands, like CREATE, then ALTER, then DROP (one-time data coverters for example).

c24t:
> `CREATE TABLE table (col1 type1)` looks like a query to me, especially if it is a big script with several commands, like CREATE, then ALTER, then DROP (one-time data converters for example).

I'm using spanner's own terms for queries and statements here according to https://cloud.google.com/spanner/docs/query-syntax and https://cloud.google.com/spanner/docs/data-definition-language. I agree that "statement" is a pretty broad term (as compared to something like "DDL statement"), but "query" has a particular meaning in spanner and other cloud db products: they have to start with a query statement, include a query expression, scan tables and return rows, etc.

IlyaFaer:
@c24t, hm-m, I've never thought of this page like a strict definition of what is "query", just a syntax explanation. But now, after rereading these pages (DML, DDL too) carefully, yes, it looks like "query" actually means a read request, and is never used in other cases (DML, DDL).  Okay, renaming back

c24t (approved):

PR #502, opened 2020-09-15
fix: Update version to 2.2a1

c24t:
Add a version file and clean up some hardcoded version strings (still more to go).

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -12,7 +12,6 @@ from setuptools import find_packages, setup
 # Package metadata.
 name = "django-google-spanner"
 description = "Bridge to enable using Django with Spanner."
-version = "2.2a0"
 # Should be one of:
 # 'Development Status :: 3 - Alpha'
 # 'Development Status :: 4 - Beta'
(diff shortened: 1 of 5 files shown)


setup.py
```diff
@@ -22,23 +21,28 @@
 dependencies = ["sqlparse >= 0.3.0", "google-cloud-spanner >= 1.8.0"]
 extras = {}
 
+BASE_DIR = os.path.dirname(__file__)
+VERSION_FILENAME = os.path.join(BASE_DIR, "version.py")
+PACKAGE_INFO = {}
+with open(VERSION_FILENAME) as f:
+    exec(f.read(), PACKAGE_INFO)
+version = PACKAGE_INFO["__version__"]
```
mf2199, line 29: This redeclares the definition in L#15 which thereby seems to have no use.

mf2199 (commented):
A small nit, otherwise looks good.

setup.py
c24t: Good catch, forgot to remove L15.

PR #503, opened 2020-09-15
fix: Update README for alpha release

c24t:
Cosmetic changes to the README for the upcoming alpha release.

Supersedes #490.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1,61 +1,50 @@
-django-spanner
-==============
+Cloud Spanner support for Django
+================================
 
-ORM plugin for using Cloud Spanner as a database for Django.
+ORM plugin for using Cloud Spanner as a `database backend
+<https://docs.djangoproject.com/en/2.2/ref/databases/#using-a-3rd-party-database-backend>`__
+for Django.
 
-🚨THIS CODE IS STILL UNDER DEVELOPMENT🚨
-========================================
 
-Table of contents
------------------
+Installation
+------------
 
--  `Installing it <#installing-it>`__
--  `Using it <#using-it>`__
+To use this library you'll need a Google Cloud Platform project with the Cloud
+Spanner API enabled. See the `Cloud Spanner Python client docs
+<https://github.com/googleapis/python-spanner/#quick-start>`__ for details.
 
-   -  `Format <#format>`__
-   -  `Example <#example>`__
+Use the version of ``python-spanner-django`` that corresponds to your version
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -1,61 +1,50 @@
...
-The minor release number of Django doesn't correspond to the minor
-release number of django-spanner. Use the latest minor release of each.
+    git clone git@github.com:googleapis/python-spanner-django.git
+    cd python-spanner-django
+    pip3 install -e .
 
-.. code:: shell
 
-    pip3 install --user .
+Use
```
mf2199, line 39: [nit] Instead of `Installing it` and `Using it` gerunds as section names, I'd rather see nouns like `Installation` and `Usage`, as more semantically aligned with `Limitations`.

README.rst
```diff
@@ -107,18 +96,18 @@ Limitations
 Transaction management isn't supported
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-django-spanner always works in Django's default transaction behavior,
-``autocommit`` mode. Transactions cannot be controlled manually with
+``python-spanner-django`` always works in Django's default transaction
```
mf2199, line 99: [nit] Maybe `The ''python-spanner-django''...`?

README.rst
```diff
@@ -1,61 +1,50 @@
-django-spanner
-==============
+Cloud Spanner support for Django
```
mf2199, line 1: Would it be better to have `Cloud Spanner support for Django framework`?

README.rst
mf2199: Also, the usage of comma after `behavior` is unclear. Maybe `...transaction behavior, the ''autocommit'' mode.` Then it's not Django's but Spanner's default transaction mode, WDYT?

c24t: I agree that's not clear, changed it to say:

> `python-spanner-django` always works in `autocommit` mode, which is Django's default behavior even for backends that support manual transaction management.

I think this is the same sentiment but more explicit.

c24t: > [nit] Maybe The ''python-spanner-django''...?

I don't think the package name needs an article.

README.rst
c24t: I think it's ok as is, django is very popular and I'd expect people looking at this package to know what it is.

README.rst
c24t: Agreed, I changed "use" and the last instance of "installing it".

PR #504, opened 2020-09-15
refactor: erase strip_backticks() helper

IlyaFaer:


google/cloud/spanner_dbapi/parse_utils.py
```diff
@@ -68,15 +68,6 @@ def classify_stmt(sql):
 )
 
 
-def strip_backticks(name):
-    """
-    Strip backticks off of quoted names For example, '`no`' (a Spanner reserved
-    word) becomes 'no'.
-    """
-    has_quotes = name.startswith("`") and name.endswith("`")
-    return name[1:-1] if has_quotes else name
```
IlyaFaer, line 77: Doesn't look like we actually need such a logic. This function is used only on column names. Column names themselves can not start or end with backticks, so backticks are used only as a marking. Means we can just `strip()` both whitespaces and backticks with one method call.

google/cloud/spanner_dbapi/parse_utils.py
```diff
@@ -169,10 +160,7 @@ def parse_insert(insert_sql, params):
     if values.homogenous():
         # Case c)
 
-        columns = [
-            strip_backticks(mi.strip())
-            for mi in match.group("columns").split(",")
-        ]
```
IlyaFaer, line 175: This `columns` list is not used in calculations, only its length. The column names themselves are only used in the "unequal length" error message

c24t (approved):
LGTM

PR #507, opened 2020-09-15
fix: Fix license classifier

c24t:
Fixes this error on `twine upload`:

```
HTTPError: 400 Bad Request from https://upload.pypi.org/legacy/
Invalid value for classifiers. Error: Classifier 'License :: OSI Approved :: BSD' is not a valid classifier.
```


mf2199 (approved):
👍  Took me some time to find a list of those.

PR #516, opened 2020-09-21
feat: avoid adding dummy WHERE clause into UPDATE and DELETE queires

IlyaFaer:


IlyaFaer:
@c24t, looks like it's our way to raise exceptions on lack of WHERE statement, but not to break all the Django tests. What do you think?

google/cloud/spanner_dbapi/parse_utils.py
```diff
@@ -402,107 +502,13 @@ def ensure_where_clause(sql):
...
-    "USING",
-    "WHEN",
-    "WHERE",
-    "WINDOW",
-    "WITH",
-    "WITHIN",
-}
+    raise ProgrammingError(
+        "Cloud Spanner requires a WHERE clause when executing DELETE or UPDATE query"
+    )
```
skuruppu, line 511: Apologies if I misunderstood something here.

I believe this clause is added by default because Django will construct  DELETE and UPDATE queries that don't have the WHERE clause as this is not a requirement in other databases. It ensure that customers can smoothly transition from their existing applications without having to write raw queries with the WHERE clause.

The clause wasn't added just to get the tests to pass.

IlyaFaer: @skuruppu, hiding this useful restriction doesn't look right. But I see the point in smooth transition as well. Do you mean that only for Django? Or we should consider that any user can have queries without WHERE, so this dummy should be kept for the whole connection API? (we have `spanner_dbapi` directory, which seems to be intended to be default connection API - it's from where I'm proposing to erase the dummy, and `django_spanner` directory, where Django-specific parts are held).

The main thing that makes me nervous is that queries can be big and branched (for example, UNIONS are little bit hard to interpret, as WHERE clause can be relative to the last of the union parts, but also can be relative to the result of the UNION on the whole, and they both will be at the about same place). I'm not sure if there is a way to check 100% of variants with several regexes, and insert this dummy correctly. So, I think it'll be exploding things from time to time.

skuruppu: I agree that it may be good to notify the user about this restriction from the connection API. Is it possible to instead modify the Django implementation to add the WHERE clause in the generated query? I think this is where it matters the most in order to support users migrating from other databases.

c24t:
Talking about this PR in person: following @skuruppu's comment, these changes should move to django-specific code.

IlyaFaer:
@c24t, I've pushed the changes, it's seems to be ready for merge now.

tests/unit/spanner_dbapi/test_parse_utils.py
```diff
@@ -407,36 +407,26 @@ def test_get_param_types_none(self):
...
+        for sql in cases:
+            with self.subTest(sql=sql):
+                ensure_where_clause(sql)
 
-        for sql, want in cases:
+        for sql in err_cases:
             with self.subTest(sql=sql):
-                got = ensure_where_clause(sql)
-                self.assertEqual(got, want)
+                with self.assertRaises(ProgrammingError):
```
c24t, line 428: Might be worth testing the same thing in `test_parser` to make sure `ensure_where_clause` is actually getting called.

django_spanner/utils.py
```diff
@@ -18,3 +19,17 @@ def check_django_compatability():
                 A=django.VERSION[0], B=django.VERSION[1], C=__version__
             )
         )
+
+
+def ensure_where_clause(sql):
```
c24t, line 24:
```suggestion
def add_dummy_where(sql):
```

Pretty confusing that this function name shows up twice, and with two different meanings of "ensure"!

c24t (approved):
LGTM, but please rename one of those `ensure_where_clause` funcs.

Note that this will have to be split up across python-spanner and spanner-django now because of https://github.com/googleapis/python-spanner/pull/160.

c24t:
@IlyaFaer just pushed some minor fixes for my own comments, if it LGTY we can merge.

c24t (approved):

IlyaFaer:
@c24t, yeah, I'm okay with this changes

PR #517, opened 2020-09-22
refactor: update and document Cursor.executemany() method, add unit t…

IlyaFaer:
…ests

### DIFF
--- a/google/cloud/spanner_dbapi/cursor.py
+++ b/google/cloud/spanner_dbapi/cursor.py
@@ -208,10 +208,7 @@ class Cursor:
         return self
 
     def __exit__(self, etype, value, traceback):
-        self.__clear()
-
-    def __clear(self):
-        self._connection = None
+        self.close()
 
     @property
     def description(self):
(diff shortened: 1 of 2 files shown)


google/cloud/spanner_dbapi/cursor.py
```diff
@@ -208,10 +208,7 @@ def __enter__(self):
         return self
 
     def __exit__(self, etype, value, traceback):
-        self.__clear()
-
-    def __clear(self):
-        self._connection = None
```
IlyaFaer, line 214: We don't actually need to drop connection object, checking `is_closed` will do the job

google/cloud/spanner_dbapi/cursor.py
```diff
@@ -208,10 +208,7 @@ def __enter__(self):
...
+        self.close()
```
IlyaFaer, line 211: On exiting `Cursor` context we should just close the cursor

mf2199 (approved):

PR #521, opened 2020-09-30
feat: Implementing DB-API types according to the PEP-0249 specification

mf2199:
The proposed changes represent reworked `types.py` in accordance with the **PEP-0249** [Type Objects and Constructors](https://www.python.org/dev/peps/pep-0249/#type-objects-and-constructors) specifications.

_Some of the classes implemented prior to the alpha release are not part of the DB-API standards. Those were left in for now but may be removed later in the GA release._

### DIFF
--- a/google/cloud/spanner_dbapi/types.py
+++ b/google/cloud/spanner_dbapi/types.py
@@ -12,6 +12,7 @@
 """
 
 import datetime
+from base64 import b64encode
 
 
 def _time_from_ticks(ticks, tz=None):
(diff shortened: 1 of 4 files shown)


google/cloud/spanner_dbapi/types.py
```diff
@@ -4,92 +4,67 @@
...
-    """
-    return b64encode(string)
-
-
-class BINARY:
-    """
-    This object describes (long) binary columns in a database (e.g. LONG, RAW, BLOBS).
-    """
+def _time_from_ticks(ticks, tz=None):
+    """A helper method used to construct a DB-API time value.
```
IlyaFaer, line 19: I think we should first add system tests for these types, before merging. As they are gonna be used to convert data from user types to the Spanner types, we better check how it's working in the case close to the real.

mf2199: @IlyaFaer Reverted to the previous convention, PTAL.

skuruppu (approved):
I think this change is reasonable and looks a lot cleaner now. Just to have another pair of eyes, I'll add @larkee as a reviewer but if this is urgent, feel free to merge.

PR #522, opened 2020-09-30
refactor: verify and comment the DB API exceptions

IlyaFaer:


mf2199 (commented):
The comments look fine, I'd just rearrange them a bit to get rid of some of the unnecessary line breaks. See the suggested variant [here](https://github.com/q-logic/python-spanner-django/pull/10/files).

google/cloud/spanner_dbapi/exceptions.py
```diff
@@ -4,42 +4,91 @@
...
 class Warning(Exception):
+    """Important DB API warning."""
+
     pass
 
 
 class Error(Exception):
+    """The base class for all the DB API exceptions.
+
+    Does not include :class:`Warning`.
```
mf2199, line 19: This might be stating the obvious and can be rephrased.

Issue #485, opened 2020-09-10
Don't use django fork in build script


c24t:
The CI build script is using @timgraham's fork of django:

https://github.com/googleapis/python-spanner-django/blob/94ba284118c6cb02fccc9d40bb6c1e52d3a532a0/.kokoro/build.sh#L60

It looks like this was added in https://github.com/googleapis/python-spanner-django/commit/27222a3418d1f86bc67ba52a49e369c4c82b9148, but I can't tell why we needed to use this instead of `django/django`. Maybe because @timgraham merged a fix on his fork before it it was available on main?

In any case we should remove this now if we can. Related to the test cleanup work in #471.

prakhargurunani:
@c24t Can I take this issue ?

c24t:
@prakhargurunani if you're willing to give it a shot, definitely! @IlyaFaer may have more detail on the reasons we can't just use mainline django now.

PR #523, opened 2020-09-30
refactor: refactor Connection() class and cover it with unit tests

IlyaFaer:
Add warnings for methods which can't be used in autocommit mode. Update `Connection()` class constructor, and cover it with unit tests.

### DIFF
--- a/django_spanner/features.py
+++ b/django_spanner/features.py
@@ -217,6 +217,7 @@ class DatabaseFeatures(BaseDatabaseFeatures):
         "admin_views.test_multidb.MultiDatabaseTests.test_change_view",
         "admin_views.test_multidb.MultiDatabaseTests.test_delete_view",
         "auth_tests.test_admin_multidb.MultiDatabaseTests.test_add_view",
+        "auth_tests.test_remote_user_deprecation.RemoteUserCustomTest.test_configure_user_deprecation_warning",
         "contenttypes_tests.test_models.ContentTypesMultidbTests.test_multidb",
         # Tests that by-pass using django_spanner and generate
         # invalid DDL: https://github.com/orijtech/django-spanner/issues/298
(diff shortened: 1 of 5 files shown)


google/cloud/spanner_dbapi/connection.py
```diff
@@ -122,19 +144,16 @@ def close(self):
...
-        self.run_prior_DDL_statements()
+        """Commit all the pending transactions."""
+        warnings.warn(AUTOCOMMIT_MODE_WARNING, UserWarning, stacklevel=2)
 
     def rollback(self):
-        self._raise_if_closed()
-
-        # TODO: to be added.
+        """Rollback all the pending transactions."""
+        warnings.warn(AUTOCOMMIT_MODE_WARNING, UserWarning, stacklevel=2)
```
IlyaFaer, line 156: I've used just `warnings` shots as raising an exception in these methods will break a lot of Django tests.

django_spanner/features.py
```diff
@@ -217,6 +217,7 @@ class DatabaseFeatures(BaseDatabaseFeatures):
         "admin_views.test_multidb.MultiDatabaseTests.test_change_view",
         "admin_views.test_multidb.MultiDatabaseTests.test_delete_view",
         "auth_tests.test_admin_multidb.MultiDatabaseTests.test_add_view",
+        "auth_tests.test_remote_user_deprecation.RemoteUserCustomTest.test_configure_user_deprecation_warning",
```
IlyaFaer, line 220: There is no such a test in the current Django version. That related to [the issue](https://github.com/googleapis/python-spanner-django/issues/485) about using an old Django fork for testing (it is 2000 behind the original Django version)

mf2199 (approved):

PR #525, opened 2020-10-01
refactor: update expect() parse function

IlyaFaer:
There is a couple of issues related to the parsing functions. Trying to reduce their complexity first before starting issues.

### DIFF
--- a/google/cloud/spanner_dbapi/parser.py
+++ b/google/cloud/spanner_dbapi/parser.py
@@ -30,7 +30,6 @@ from .exceptions import ProgrammingError
 ARGS = "ARGS"
 EXPR = "EXPR"
 FUNC = "FUNC"
-TERMINAL = "TERMINAL"
 VALUES = "VALUES"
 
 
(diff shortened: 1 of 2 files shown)


google/cloud/spanner_dbapi/parser.py
```diff
@@ -150,38 +149,29 @@ def parse_values(stmt):
 
 
 def expect(word, token):
-    word = word.strip(" ")
+    word = word.strip()
```
IlyaFaer, line 152: `strip()` erases whitespaces by default

google/cloud/spanner_dbapi/parser.py
```diff
@@ -150,38 +149,29 @@ def parse_values(stmt):
...
 
             word, arg = expect(word, ARGS)
             all_args.append(arg)
+            word = word.strip()
 
-            word = word.strip(" ")
-            if word == "":
-                break
-            elif word[0] != ",":
+            if word and not word.startswith(","):
```
IlyaFaer, line 168: As `while` circle already checks that `word` is not empty, we can reduce number of conditions here

google/cloud/spanner_dbapi/parser.py
```diff
@@ -150,38 +149,29 @@ def parse_values(stmt):
...
-                word = word[1:]
+            word = word[1:]
         return "", values(all_args)
 
-    elif token == TERMINAL:
-        word = word.strip(" ")
-        if word != "%s":
-            raise ProgrammingError("TERMINAL: `%s` is not %%s" % word)
-        return "", pyfmt_str
-
```
IlyaFaer, line 184: I didn't find any call to `expect` with `TERMINAL` token, so it looks unused

tests/spanner_dbapi/test_parser.py
```diff
@@ -21,35 +20,6 @@
...
-            ("%sa", "TERMINAL: `%sa` is not %s"),
-        ]
-
-        for text, wantException in cases:
-            with self.subTest(text=text):
-                self.assertRaisesRegex(
-                    ProgrammingError,
-                    wantException,
-                    lambda: expect(text, TERMINAL),
-                )
```
IlyaFaer, line 51: As `TERMINAL` token doesn't seem to be used anywhere, we don't need tests for it

mf2199 (approved):

PR #527, opened 2020-10-05
test: don't use Django fork for testing

### LINKED ISSUE #485: Don't use django fork in build script

IlyaFaer:
Closes #485

### DIFF
--- a/.kokoro/build.sh
+++ b/.kokoro/build.sh
@@ -57,7 +57,7 @@ pip3 install .
 # Create a unique DJANGO_TESTS_DIR per worker to avoid
 # any clashes with configured tests by other workers.
 export DJANGO_TESTS_DIR="django_tests_$DJANGO_WORKER_INDEX"
-mkdir -p $DJANGO_TESTS_DIR && git clone --depth 1 --single-branch --branch spanner-2.2.x https://github.com/timgraham/django.git $DJANGO_TESTS_DIR/django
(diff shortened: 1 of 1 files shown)


IlyaFaer:
@busunkim96, could you take a look and maybe advise something? I saw you dealing with similar problems.

According to [the logs](https://source.cloud.google.com/results/invocations/0ffc3279-2544-4255-8df5-3d3f34b2cc25/targets/cloud-devrel%2Fclient-libraries%2Fpython%2Fgoogleapis%2Fpython-spanner-django%2Fpresubmit%2Fworker_3/log) `requests` package is already installed, but somewhy the code can't import it. I've tried to add package update into several places, but it doesn't give any effect. Seems like something environment related to me.

Thanks in advance!

busunkim96:
Hi @IlyaFaer,

I would create a new virtual environment and install the test requirements there. It's possible to force an upgrade of the requests version installed in the kokoro environment, but Kokoro may be depending on the specific version installed.

```
python3 -m venv env
source env/bin/activate
python3 -m pip install -r requirements.txt ...
```

PR #535, opened 2020-10-12
feat: support transactions management

IlyaFaer:


google/cloud/spanner_dbapi/connection.py
```diff
@@ -37,11 +34,46 @@ class Connection:
...
+
+    @autocommit.setter
+    def autocommit(self, value):
+        """Change this connection autocommit mode.
+
+        :type value: bool
+        :param value: New autocommit mode state.
+        """
+        if value and not self._autocommit:
+            self.commit()
```
IlyaFaer, line 66: In accordance with Java implementation - in PEP 249 there is no recommendations for this

google/cloud/spanner_dbapi/cursor.py
```diff
@@ -88,6 +89,20 @@ def execute(self, sql, args=None):
...
+                or self.transaction.committed
+                or self.transaction.rolled_back
+            ):
+                self.transaction = self._connection.session_checkout().transaction()
+                self.transaction.begin()
+                self._connection.transactions.append(self.transaction)
+
+            self._res = self.transaction.execute_sql(sql)
+            self._itr = PeekIterator(self._res)
+            return
```
IlyaFaer, line 104: I no transaction started for this cursor, or the last transaction is commited/rollbacked, we're starting new transaction. If transaction is in progress, we're just running the given SQL in it.

tests/system/test_system.py
```diff
@@ -6,17 +6,281 @@
...
+
+        Config.INSTANCE_CONFIG = configs[0]
+        config_name = configs[0].name
+
+        Config.INSTANCE = Config.CLIENT.instance(INSTANCE_ID, config_name)
+        created_op = Config.INSTANCE.create()
+        created_op.result(30)  # block until completion
+    else:
+        Config.INSTANCE = Config.CLIENT.instance(INSTANCE_ID)
+        Config.INSTANCE.reload()
```
IlyaFaer, line 97: All of the above code is taken from the original Spanner system tests

IlyaFaer:
@olavloite, @c24t, I've written the base implementation for transactions management. All four system tests are passing locally, though sometimes transactions are aborted (this is okay until we're not implemented an aborted transactions retry, I suppose). User interface is completely the same as other Python databases, e.g. [SQLite 3](https://docs.python.org/3/library/sqlite3.html), so I suppose we're suiting PEP 249 completely.

tests/system/test_system.py
```diff
@@ -6,17 +6,281 @@
...
+SET first_name = 'updated-first-name'
+WHERE first_name = 'first-name'
+"""
+        )
+        conn.autocommit = True
+
+        # read the resulting data from the database
+        cursor.execute("SELECT * FROM contacts")
+        got_rows = cursor.fetchall()
+        conn.commit()
```
olavloite, line 241: This should not be necessary (and now probably causes a warning), as the connection is in autocommit mode.

google/cloud/spanner_dbapi/connection.py
```diff
@@ -37,11 +34,46 @@ class Connection:
     """
 
     def __init__(self, instance, database):
+        self._pool = BurstyPool()
```
olavloite, line 37: Having a connection pool for each `Connection` is too much. One connection will never use more than one session at any time, as:
1. A connection can execute at most one transaction at any time.
1. A session can execute at most one (read/write) transaction at any time.

This is fine for a prototype, for production purposes, the implementation should do one of the following:
1. Create one session pool globally that is reused for all connections that are opened. This is the preferable implementation, as the connections would benefit from all optimizations that might be done to the session pool implementation.
1. OR: Open one session (not session pool) for each connection, and use that session for all operations on that connection. Conceptually, a Cloud Spanner session is quite comparable to a Python database connection. Implementing it this way would however mean that the connection needs to do all session management by itself, including things like keep-alive pings, handling the backend dropping a session, etc. It is therefore not recommended to do that, and instead rely on the session pool handling all that for you.

google/cloud/spanner_dbapi/connection.py
```diff
@@ -37,11 +34,46 @@ class Connection:
...
+
+        self._autocommit = value
+
+    def session_checkout(self):
+        """Get a Cloud Spanner session.
+
+        :rtype: :class:`google.cloud.spanner_v1.session.Session`
+        :returns: Cloud Spanner session object ready to use.
+        """
+        return self._pool.get()
```
olavloite, line 76: It seems that this session will never be returned to the pool, which will cause the connection to leak a session each time a transaction is executed.

google/cloud/spanner_dbapi/cursor.py
```diff
@@ -88,6 +89,20 @@ def execute(self, sql, args=None):
 
         self._res = None
 
+        if not self._connection.autocommit:
+            if (
+                not self.transaction
+                or self.transaction.committed
+                or self.transaction.rolled_back
+            ):
+                self.transaction = self._connection.session_checkout().transaction()
```
olavloite, line 98: The session that is checked out here, will never be returned to the pool. It is probably better to keep a reference to the current session and transaction in the `Connection` class, and to return the session to the pool when the active transaction on the connection is committed or rolled back.

olavloite:
@IlyaFaer Thanks for the draft. This seems like a good first step towards transaction management.

One important design decision to made here, is how we should handle sessions. Preferably, the Connection API should rely on the session pool implementation from the client library. That would prevent us from having to implement all session handling once more for connections. It does however mean that there must be some kind of global session pool that is reused by all connections that are opened.

google/cloud/spanner_dbapi/connection.py
```diff
@@ -142,18 +174,34 @@ def get_table_column_schema(self, table_name):
...
+            transaction.rollback()
+
         self.__dbhandle = None
         self.is_closed = True
 
     def commit(self):
         """Commit all the pending transactions."""
-        warnings.warn(AUTOCOMMIT_MODE_WARNING, UserWarning, stacklevel=2)
+        if self.autocommit:
+            warnings.warn(AUTOCOMMIT_MODE_WARNING, UserWarning, stacklevel=2)
```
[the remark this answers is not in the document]
IlyaFaer: Can't say I completely agree on that. If one is calling `commit()` in `autocommit=true` mode, it's definitely a logical error, but I don't think it'll break data or current operations flow. So, as I see it, we can, you know, forgive it, but warningly wave a finger.

In Java you're throwing an error in such a cases?

google/cloud/spanner_dbapi/__init__.py
```diff
@@ -93,7 +97,7 @@ def connect(
     if not database.exists():
         raise ValueError("database '%s' does not exist." % database_id)
 
-    return Connection(instance, database)
+    return Connection(instance, database, default_pool)
```
IlyaFaer, line 100: While in standard interface, we're putting the default pool. However, one will be able to set his own pool by initiating a `Connection` object manually.

google/cloud/spanner_dbapi/connection.py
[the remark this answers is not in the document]
olavloite: I would recommend that we follow the convention of other Python DB API implementations. So the most important question is what do the implementations for for example PostgreSQL and MySQL do?

The Java implementation throws an exception if you try to call `commit` or `rollback` when in autocommit mode.

google/cloud/spanner_dbapi/__init__.py
olavloite: That sounds good.

Could we add an exit handler that automatically closes the default pool (if it exists and is open) when the application is shutdown?

google/cloud/spanner_dbapi/__init__.py
```diff
@@ -47,6 +47,10 @@
 # threadsafety level.
 threadsafety = 1
 
+# Cloud Spanner sessions pool, used by
+# default for the whole DB API package
+default_pool = spanner_v1.BurstyPool()
```
olavloite, line 52: Could we make this lazy and only initialize it the first time a connection is opened? Opening a session pool can be an expensive operation (depending on what the default settings for the pool are).

google/cloud/spanner_dbapi/connection.py
```diff
@@ -34,14 +31,87 @@ class Connection:
...
+            self._session = self._pool.get()
+
+        return self._session
+
+    def _release_session(self):
+        """Release the currently used Spanner session.
+
+        The session will be returned into the sessions pool.
+        """
+        self._pool.put(self._session)
```
olavloite, line 93: What happens if `self._session` is `None` here?

tests/system/test_system.py
```diff
@@ -6,17 +6,292 @@
 
 import unittest
 
+# Copyright 2016 Google LLC All rights reserved.
```
olavloite, line 9: nit: 2020

google/cloud/spanner_dbapi/__init__.py
IlyaFaer: I think it is lazy (at least, it doesn't post requests to a backend). In the original `BurstyPool` class init, it only creates an empty queue of the given size. And in `bind()` method (which will be called on `Connection` init) it also just saves a pointer to the database object:
https://github.com/googleapis/python-spanner/blob/39ea9489475619469044573c346349b748e28d4b/google/cloud/spanner_v1/pool.py#L247-L260

In the base init it also doesn't do anything very meaningful, only initiates a labels dictionary:
https://github.com/googleapis/python-spanner/blob/39ea9489475619469044573c346349b748e28d4b/google/cloud/spanner_v1/pool.py#L38-L41

Only when we'll try to get a session (on a transaction begin), the pool will start to actually do something:
https://github.com/googleapis/python-spanner/blob/39ea9489475619469044573c346349b748e28d4b/google/cloud/spanner_v1/pool.py#L269-L273

google/cloud/spanner_dbapi/connection.py
IlyaFaer: It's not gonna come here when `self._session` is `None`, as this method is only called when we have a `self._transaction is not None`; and there is no way to get a transaction without session

google/cloud/spanner_dbapi/connection.py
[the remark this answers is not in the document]
IlyaFaer: It looks like every database solves this question by itself. In [SQLite3](https://github.com/coleifer/pysqlite3/blob/master/src/connection.c) I see they are doing nothing if someone called `commit()` in autocommit mode (not even throwing a warning). [MySQL](https://github.com/mysql/mysql-connector-python/blob/5dba2597d9c52f348ee1a3b0b6628f2c035cdc9e/lib/mysql/connector/connection.py) just sends a `COMMIT` statement without any checks. [Postgres](https://github.com/psycopg/psycopg2/blob/f2852a520a26d88fb248c0563e73af43110a13a8/psycopg/pqpath.c) connector prints a warning `no transaction to commit`.

Thus, plus-minus, in Python we usually don't fail on calling commit in autocommit mode. I suppose it's not the mode users often changing, so the chance that one will make a mistake here is low.

google/cloud/spanner_dbapi/__init__.py
IlyaFaer: This can be done with [atexit](https://docs.python.org/3/library/atexit.html) package. That's the question to ask googlers I suppose, as I'm not sure if there are any underwater rocks.

IlyaFaer:
@skuruppu, @larkee, PTAL, transactions management seems to be more or less implemented. We still have some ongoing conversations with @olavloite, but I think it'll be good if you'll take a look as well, and evaluate the concepts (with Python expertize too).

IlyaFaer:
@olavloite, @c24t, I've re-read the JDBC design doc - it looks like only DDL statements should not be executed within transaction. Fixed this part and added the args parsing (using the existing functions).

I've also fixed emulator configurations, so system tests on emulator are now working fine.

google/cloud/spanner_dbapi/connection.py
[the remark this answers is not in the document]
c24t: If we didn't care about following DBAPI conventions I'd agree this should be an error. But sqlite3 is the oldest and likely most-used DBAPI library, and IIRC used as the reference implementation for DBAPI. Following its lead, I think it's fine for this and `rollback` to warn instead of raise.

google/cloud/spanner_dbapi/connection.py
```diff
@@ -142,18 +174,34 @@ def get_table_column_schema(self, table_name):
...
+            for transaction in self.transactions:
+                transaction.commit()
+
+            self.transactions = []
 
     def rollback(self):
         """Rollback all the pending transactions."""
-        warnings.warn(AUTOCOMMIT_MODE_WARNING, UserWarning, stacklevel=2)
+        if self.autocommit:
+            warnings.warn(AUTOCOMMIT_MODE_WARNING, UserWarning, stacklevel=2)
```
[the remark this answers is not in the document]
c24t: Same as above, I think we ought to follow sqlite here.

google/cloud/spanner_dbapi/__init__.py
c24t: I don't see any reason not to use atexit here -- just that the handler should include a short timeout so we don't keep the process open if we don't get a timely response from spanner.

google/cloud/spanner_dbapi/__init__.py
c24t: We do run the risk of the `BurstyPool` implementation changing, or some future contributor changing this to another pool type with an expensive initialization. If that happens I could see users being very surprised that just importing this module causes API calls to spanner. Considering it's an easy change to make, I agree that it's a good idea to make this lazy.

c24t (commented):
A few comments, but it looks very good overall. I think this could reasonably be approved as-is, but @IlyaFaer I'll give you a chance to respond before I stamp it.

Also thanks @olavloite for the detailed review!

google/cloud/spanner_dbapi/connection.py
```diff
@@ -34,14 +31,125 @@ class Connection:
...
+
+    @database.setter
+    def database(self, _):
+        """Related database setter.
+
+        Restricts replacing the related database.
+        """
+        raise ProgrammingError(
+            "Replacing the related database of the existing connection is restricted."
+        )
```
mf2199, line 91: Why not just leave the setter out and make the property "read-only"?

google/cloud/spanner_dbapi/connection.py
```diff
@@ -34,14 +31,125 @@ class Connection:
...
+
+    @instance.setter
+    def instance(self, _):
+        """Related instance setter.
+
+        Restricts replacing the related instance.
+        """
+        raise ProgrammingError(
+            "Replacing the related instance of the existing connection is restricted."
+        )
```
mf2199, line 110: Same as for the `database` setter.

google/cloud/spanner_dbapi/connection.py
```diff
@@ -34,14 +31,125 @@ class Connection:
...
+
+    def _session_checkout(self):
+        """Get a Cloud Spanner session from a pool.
```
mf2199, line 113: `...the pool.`

google/cloud/spanner_dbapi/connection.py
```diff
@@ -34,14 +31,125 @@ class Connection:
...
+        The session will be returned into the sessions pool.
+        """
+        self._pool.put(self._session)
+        self._session = None
+
+    def transaction_checkout(self):
+        """Get a Cloud Spanner transaction.
+
+        Begin a new transaction, if there is no transaction in
+        this connection yet. Return the begun one otherwise.
```
mf2199, line 138: Rather, `the existing one`.

google/cloud/spanner_dbapi/connection.py
```diff
@@ -34,14 +31,125 @@ class Connection:
...
+
+        :rtype: :class:`google.cloud.spanner_v1.transaction.Transaction`
+        :returns: Cloud Spanner transaction object ready to use.
```
mf2199, line 141: `A Cloud Spanner transaction object, ready to use.`

google/cloud/spanner_dbapi/connection.py
mf2199: @IlyaFaer That is true for the code at present. However, considering Murphy Law with possible future upgrades, a simple check may be warranted.

larkee (changes requested):
There seems to be confusion about session pools and database binding. I've tried to explain best I can but let me know if you have further questions :+1:

google/cloud/spanner_dbapi/__init__.py
```diff
@@ -93,7 +96,17 @@ def connect(
     if not database.exists():
         raise ValueError("database '%s' does not exist." % database_id)
 
-    return Connection(instance, database)
+    if config.default_pool is None:
+        config.default_pool = spanner_v1.BurstyPool()
```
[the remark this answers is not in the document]
c24t: > These lines seem to do the same thing that is already achieved in this line in the `Connection` class:

I think those were meant to be removed in efbdab3, good catch.

> Given that the user is passing in the `Database` object, is it reasonable to instruct them to set up their pool requirements when creating it?

AFAICT this is all to get a reference to pool from the connection class. I agree it's better to use `database._pool` (or even better: make `_pool` public or expose methods for get/putting sessions) than wire it through like this.

google/cloud/spanner_dbapi/connection.py
```diff
@@ -34,14 +31,125 @@ class Connection:
...
+    :param pool: (Optional) Cloud Spanner sessions pool.
     """
 
-    def __init__(self, instance, database):
-        self.instance = instance
-        self.database = database
-        self.is_closed = False
+    def __init__(self, instance, database, pool=None):
+        self._pool = pool or BurstyPool()
+        self._pool.bind(database)
```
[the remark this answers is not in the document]
c24t: Or remove `_pool` altogether and replace it with `self.database._pool` in `_session_checkout` and `_release_session` below.

google/cloud/spanner_dbapi/connection.py
```diff
@@ -34,14 +31,87 @@ class Connection:
...
         self.instance = instance
         self.database = database
-        self.is_closed = False
 
         self._ddl_statements = []
+        self._transaction = None
+        self._session = None
+
+        self.is_closed = False
+        self._autocommit = False
```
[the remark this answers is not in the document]
c24t: That sounds fine to me, I just mean we have to respect django's autocommit setting here.

google/cloud/spanner_dbapi/config.py
```diff
@@ -0,0 +1,12 @@
...
+# Use of this source code is governed by a BSD-style
+# license that can be found in the LICENSE file or at
+# https://developers.google.com/open-source/licenses/bsd
+
+"""Configurations module for DB API."""
+
+# Cloud Spanner sessions pool, used by
+# default for the whole DB API package.
+# Is lazily initiated on a connection creation.
+default_pool = None
```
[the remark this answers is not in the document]
IlyaFaer: It's written in accordance with Python recommendation: https://docs.python.org/3/faq/programming.html#how-do-i-share-global-variables-across-modules

google/cloud/spanner_dbapi/connection.py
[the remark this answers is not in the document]
IlyaFaer: Got it, pushing. In fact the first version of transaction management wasn't using a global pool, so it's not gonna be a problem to return to it.

c24t (approved):
LGTM, @larkee any comments before merging this?

google/cloud/spanner_dbapi/connection.py
```diff
@@ -37,11 +33,96 @@ class Connection:
...
+        if not self.autocommit:
+            if (
+                not self._transaction
+                or self._transaction.committed
+                or self._transaction.rolled_back
+            ):
+                self._transaction = self._session_checkout().transaction()
+                self._transaction.begin()
+
+            return self._transaction
```
larkee, line 125: Just to confirm, the transaction is only returned if `autocommit` is not set. If `autocommit` is not set then nothing is returned. Is this the intended behavior? If so, please document it. If the implementation for `autocommit=True` is to come later, then this is fine :+1:

larkee (commented):
LGTM, just one question for the `transaction_checkout` method

larkee (approved):

PR #536, opened 2020-10-13
test: run system tests against emulator

IlyaFaer:
Adding a workflow file to add an ability to run tests against Spanner emulator.
Django tests should be moved into nox system tests, so that we could run them on emulator.

### DIFF
--- a/.github/workflows/system-tests-against-emulator.yaml
+++ b/.github/workflows/system-tests-against-emulator.yaml
@@ -1,0 +1,32 @@
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+name: Run Spanner integration tests against emulator
+jobs:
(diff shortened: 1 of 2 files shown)


olavloite (commented):
I have no specific comments on the changes itself. The build errors seems not to be related to this change, but to resource problems on the test project. Is that something that is fixable, either in code or in the test project?

IlyaFaer:
@olavloite, yeah, that's a flaky thing. I'm trying to fix it from several directions, by connecting to emulator, using more actual Django branch, etc. Don't have big results so far, as my local environment is far away from the one of kokoro, and I have to push and wait for results every time.

mf2199 (approved):
LGTM

PR #543, opened 2020-10-23
feat: clear session pool on connection close

IlyaFaer:
To avoid keeping Cloud Spanner sessions alive after connection close we should clear the related pool on the connection close.


google/cloud/spanner_dbapi/connection.py
```diff
@@ -235,6 +235,7 @@ def close(self):
         ):
             self._transaction.rollback()
 
+        self.database._pool.clear()
```
olavloite, line 238: If a user has created multiple connections for the same database, then I think this will invalidate the session pool of all those connections, while he/she is only closing this connection.

IlyaFaer: Yeah, I had a thought about that. To achieve this one will need to create a pool manually and propagate it to several `connect()` calls. Is it a correct way of use? I mean, what prevents user from trying to propagate one pool (binded to a single database) onto connections to several different databases.

Every `Database` in Python initiates a pool for itself (if non given), which makes me thing it's not intended to be used with global pools.

olavloite: Yeah, I should have looked a little bit more in detail into the `connection.py` file. The current implementation requires a user to pass in a `database` when a connection is created. That basically means that we are asking the user of the library to initialize and maintain a database reference themselves. That again means that the `Connection` class should not open or close any session pools for the `database`. Instead, that is the responsibility of the caller that has already obtained a reference to the `database`.

olavloite (changes requested):

c24t (commented):
I agree with @olavloite in https://github.com/googleapis/python-spanner-django/pull/543/files#r510984476 that this shouldn't clear the session pool if it doesn't own it.

FWIW in python-spanner `Database` doesn't expose the session pool or give users an option to `clear` it (unless they created the pool themselves). Like https://github.com/googleapis/python-spanner-django/pull/535#discussion_r509023513 this might be an argument for changes to python-spanner instead of this library.

IlyaFaer:
I still don't see how one can use single pool for several databases. Pool is binded to a concrete database, if we'll use the same pool for connections to different databases, it's not gonna be correct. Thus, for me it looks like users should not use global pools at all, even for several connections to the same database, as we're not very able to control whether the pool is used only for single database.

olavloite:
> I still don't see how one can use single pool for several databases. Pool is binded to a concrete database, if we'll use the same pool for connections to different databases, it's not gonna be correct. Thus, for me it looks like users should not use global pools at all, even for several connections to the same database, as we're not very able to control whether the pool is used only for single database.

Yes, that is correct, but I'm not sure that is what @c24t meant. So to summarize:
1. A session pool is as @IlyaFaer noted always linked with one database. It is not possible to have a single session pool for multiple different databases.
1. The current implementation of the DB API requires the user to supply a `database` instance when a connection is created. That means that the `connection` is not the owner of the `database` instance, and the `connection` should therefore not close or clear the session pool when the connection is closed. That is an action that the client application should take. We should add this to the documentation so it's clear for users what we expect from them.
1. Users should **not** create a new `database` instance for each `connection` they want to open. Instead, connections for the same database should use the same `database` instance. This will effectively allow multiple connections to share the same session pool. This should also be documented.
1. Assuming that we will be supporting SQLAlchemy in the future: SQLAlchemy expects users to create connections by [specifying a connection string](https://docs.sqlalchemy.org/en/13/core/connections.html) in a very similar way as connections are created in for example JDBC or ADO.NET. Once we support that, we will need to implement some kind of 'database pool' to ensure that when a user creates multiple connections to the same database, we re-use the database instance that we have already created for a previous connection for the same database.

My comment on using something like a global session pool was aimed at that last point from the list above, where 'global' means 'global per database'. So if a user creates multiple connections to different databases, we also need multiple session pools (one for each unique database).

IlyaFaer:
@olavloite, our main way to create a connection is actually [connect function](https://github.com/googleapis/python-spanner-django/blob/927e17846398a111e3cd5bfe878d5497eded0609/google/cloud/spanner_dbapi/connection.py#L201), and it works with string ids of instance and database. However, one can send a session pool into this function. If user not setting `pool`, it'll be created by the original Spanner `Database` object [here](https://github.com/googleapis/python-spanner/blob/39ea9489475619469044573c346349b748e28d4b/google/cloud/spanner_v1/database.py#L107-L111).

Thus, in fact user *can* use global pools, but looking at how `Database` [works with the given pool](https://github.com/googleapis/python-spanner/blob/39ea9489475619469044573c346349b748e28d4b/google/cloud/spanner_v1/database.py#L107-L111), I'm starting to think that the original Spanner client doesn't consider using global pools - it'll bind the given pool to the database without checking if this pool is already binded to another database. I mean, in the original client user can propagate one pool to several different databases, it'll be incorrect, but we don't think about it (it looks like we don't). So, for me, it seems like `Database` intended to use its own pool. Still, one can pass the concrete pool **to set the right type of a pool** - not to implement one global pool for several `Database` objects. PTAL at the [docs](https://googleapis.dev/python/spanner/latest/advanced-session-pool-topics.html) - they also describe only different types of pools, not global ones. That's why I consider closing pool by the connection as a right way of things. WDYT?

As the original client doesn't care about closing a binded pool of the database, I don't think this PR is very major. If we don't have a good and simple way to close the pool, I'm okay to just close the PR.

olavloite:
@IlyaFaer I would say that it would normally be a good practice to let the owner of the pool decide when a pool is closed. So that would mean:
1. If the client application passes in a pool to a database or a connection, then the client application is the owner of the pool and should close it when appropriate. If we were to let the connection close such a session pool, we might cause unexpected errors if the same session pool is used for multiple connections.
1. If the client application does not pass in a pool to a database or a connection, then the database/connection creates its own session pool. That session pool is owned by the database/connection, and should be closed by the database/connection whenever the database/connection is closed. The drawback of this solution for connections is that a connection will never really need more than one session, so it's quite an overkill to create a separate session pool for each connection.

A global session pool that can be used for different backend databases is as you say not possible, as a session pool is always bound to a specific database on the backend. What I meant with a global session pool is a global session pool **per database**. That would basically mean a pool of session pools with the database-id as the key for getting a session pool from this 'session pool pool'. That is something that would be preferable in the end for the SQLAlchemy implementation.

I had a closer look at the current `database` and `sessionpool` implementations in the Spanner Python client library, and they do a lot less than I had expected... For now I would say that we should do the following:
1. When a user creates a connection and thereby specifies a pool that should be used (i.e. the client application passes in an actual session pool): When closing the connection we should do nothing.
1. When a user creates a connection without a specific pool: Create a new session pool and use that to create the database instance. When the connection is closed, we should also close the session pool that was created when the connection was created.

IlyaFaer:
@olavloite, the thing which I don't like in our case is that we're initiating `Database()` object by `connect()` function, and then we're passing it into `Connection()` object. So, `Connection()` object getting `Database()` object with initiated pool, and it doesn't know how this database got the pool - created its own or used the one passed into constructor. We could expose `Connection()` object constructor argument or a class property to let `Connection()` object know about pool, but it looks not very beautiful for me.

What if we'll expose a `clear_pool` arg in `Connection.close()` method, which will be `True` by default? If user wants to use global pool and control its close by himself, he'll have to call close like this: `connection.close(clear_pool=False)`.

olavloite:
> @olavloite, the thing which I don't like in our case is that we're initiating `Database()` object by `connect()` function, and then we're passing it into `Connection()` object. So, `Connection()` object getting `Database()` object with initiated pool, and it doesn't know how this database got the pool - created its own or used the one passed into constructor. We could expose `Connection()` object constructor argument or a class property to let `Connection()` object know about pool, but it looks not very beautiful for me.
> 
> What if we'll expose a `clear_pool` arg in `Connection.close()` method, which will be `True` by default? If user wants to use global pool and control its close by himself, he'll have to call close like this: `connection.close(clear_pool=False)`.

I don't think we should add a parameter to the `Connection.close()` function for that. I don't think it would be clear for most users what that would mean. I would prefer a solution where the decision is automatically made by the `Connection` based on what was passed in to the `connect()` method. Can't we add a private field to the `Connection` class that indicates whether it owns the pool or not, and use that to determine whether to automatically close the pool when the connection is closed? That field can be set by the `connect()` method when the connection is created.

IlyaFaer:
> Can't we add a private field to the Connection class that indicates whether it owns the pool or not, and use that to determine whether to automatically close the pool when the connection is closed? That field can be set by the connect() method when the connection is created.

Of course we can. Can't say I like it, but seems like there is no ideal solution in here, so I'm pushing a protected property.

tests/unit/spanner_dbapi/test_connection.py
```diff
@@ -239,9 +244,7 @@ def test_run_prior_DDL_statements(self):
                 connection.run_prior_DDL_statements()
 
     def test_context(self):
-        from google.cloud.spanner_dbapi import Connection
-
-        connection = Connection(self.INSTANCE, self.DATABASE)
+        connection = self._make_connection()
```
IlyaFaer, line 247: This test has an error: `self.DATABASE` is actually a string, not a `Database` object, so passing it into `Connection` object constructor will cause errors.

olavloite:
> > Can't we add a private field to the Connection class that indicates whether it owns the pool or not, and use that to determine whether to automatically close the pool when the connection is closed? That field can be set by the connect() method when the connection is created.
> 
> Of course we can. Can't say I like it, but seems like there is no ideal solution in here, so I'm pushing a protected property.

Thank you. I think we should look some more into this later if we start adding support for SqlAlchemy. From what I understand from the documentation on that, a connection is created by SqlAlchemy itself based on a connection string, which means that we probably need to do some more management of databases and session pools.

olavloite (approved):
The change itself looks good to me.

Can we do anything with the build failures, or are those related to the general Kokoro / test problems?

IlyaFaer:
> Can we do anything with the build failures, or are those related to the general Kokoro / test problems?

Yes, system tests on Kokoro are broken because of parallelization. I'm checking all other sessions, but system tests for now only can be run succesfully on emulator (separate GitHub Workflow check).

IlyaFaer:
@c24t, we've processed this PR. Do you have any final comments?

c24t (approved):

PR #544, opened 2020-10-23
feat: support aborted transactions internal retry

### LINKED ISSUE #539: Transaction retry
Retry closed/aborted transactions in a way transparent to the user.

IlyaFaer:
Implement aborted transactions retry mechanism.

While executing SQL statements in `!autocommit` mode, connection must remember every executed statement. In case the transaction aborted, all of these statements should be re-executed. Doing this, connection also must calculate checksum of every statement results, so that we could check if the retried transaction got the same results that the original one got. In case the checksums are not equal there is no way to continue transaction due to underlying data being changed during retry.

Closes #539


google/cloud/spanner_dbapi/checksum.py
```diff
@@ -0,0 +1,72 @@
...
+    :type retried: :class:`~google.cloud.spanner_v1.transaction.ResultsChecksum`
+    :param retried: results checksum of the retried transaction.
+
+    :raises: :exc:`RuntimeError` in case if checksums are not equal.
+    """
+    if original is not None:
+        if len(retried) == len(original) and retried != original:
+            raise RuntimeError(
+                "The underlying data being changed while retrying an aborted transaction."
+            )
```
IlyaFaer, line 72: This file was just moved from the original [PR](https://github.com/q-logic/python-spanner/pull/1). Most likely some parts will be shortened, but it requires to implement the rest first for sure.

google/cloud/spanner_dbapi/connection.py
```diff
@@ -178,6 +182,42 @@ def run_prior_DDL_statements(self):
...
+            "params": params,
+            "param_types": param_types,
+            "checksum": ResultsChecksum(),
+        }
+        self._statements.append(statement)
+
+        return (
+            transaction.execute_sql(sql, params, param_types=param_types),
+            statement["checksum"],
+        )
```
IlyaFaer, line 219: As we need to track statements executed in the transaction, and several cursors can run statements in the same transaction from different points, I'm proposing to add a method into `Connection` class. It will be saving all the executed statements in the connection attribute.

In the same time, results are fetched on cursors, so it makes sense to pass a results streaming iterator as well as a pointer to a checksum back into cursor. So when a cursor is fetching the next row, it also updates a results checksum of the statement.

google/cloud/spanner_dbapi/connection.py
```diff
@@ -252,6 +293,7 @@ def rollback(self):
         elif self._transaction:
             self._transaction.rollback()
             self._release_session()
+            self._statements = []
```
IlyaFaer, line 296: When transaction is finished (commited or roll backed) we can drop all the held statements.

google/cloud/spanner_dbapi/cursor.py
```diff
@@ -67,6 +67,8 @@ def __init__(self, connection):
         self._row_count = _UNSET_COUNT
         self._connection = connection
         self._is_closed = False
+        # the currently running SQL statement results checksum
+        self._checksum = None
```
IlyaFaer, line 71: Fetching is consistent for the concrete statement, so we can use just a pointer to the current statement checksum in cursor.

google/cloud/spanner_dbapi/cursor.py
```diff
@@ -335,7 +340,9 @@ def fetchmany(self, size=None):
         items = []
         for i in range(size):
             try:
-                items.append(tuple(self.__next__()))
```
IlyaFaer, line 338: Can't say why `tuple` was used here. It doesn't make sense to me, so I assume it was just a logical error: in PEP 249 they mention returning tuples in this method as an example. But it is just an example.

IlyaFaer:
@c24t, @olavloite, I've pushed a part of transaction retry mechanism implementation.

google/cloud/spanner_dbapi/checksum.py
```diff
@@ -0,0 +1,72 @@
...
+
+    :type retried: :class:`~google.cloud.spanner_v1.transaction.ResultsChecksum`
+    :param retried: results checksum of the retried transaction.
+
+    :raises: :exc:`RuntimeError` in case if checksums are not equal.
+    """
+    if original is not None:
+        if len(retried) == len(original) and retried != original:
+            raise RuntimeError(
+                "The underlying data being changed while retrying an aborted transaction."
```
olavloite, line 71:
```suggestion
                "The transaction was aborted and could not be retried due to a concurrent modification."
```

google/cloud/spanner_dbapi/checksum.py
```diff
@@ -0,0 +1,72 @@
...
+    :param original: results checksum of the original transaction.
+
+    :type retried: :class:`~google.cloud.spanner_v1.transaction.ResultsChecksum`
+    :param retried: results checksum of the retried transaction.
+
+    :raises: :exc:`RuntimeError` in case if checksums are not equal.
+    """
+    if original is not None:
+        if len(retried) == len(original) and retried != original:
+            raise RuntimeError(
```
olavloite, line 70: I don't know the error model of Python very well, but it would be preferable if this error would be a sub class of the error that is thrown when Cloud Spanner aborts a transaction. That will make it easier for users of the API to recognize this as an `Aborted` error.

See for example the Java class [AbortedDueToConcurrentModificationException](https://github.com/googleapis/java-spanner/blob/eddd6ad4e5093ee21290b85f15fa432d071bae59/google-cloud-spanner/src/main/java/com/google/cloud/spanner/AbortedDueToConcurrentModificationException.java#L29).

google/cloud/spanner_dbapi/cursor.py
```diff
@@ -305,14 +305,19 @@ def fetchone(self):
         self._raise_if_closed()
 
         try:
-            return next(self)
+            res = next(self)
```
olavloite, line 308: I assume that the current implementation is mainly for adding the checksum feature, and not any retry logic.

But once we start adding retry logic, it is important to realize that this call to `next(self)` can cause an `Aborted` error which then again should initiate a transaction retry.

google/cloud/spanner_dbapi/cursor.py
```diff
@@ -305,14 +305,19 @@ def fetchone(self):
...
+            self._checksum.consume_result(res)
+            return res
         except StopIteration:
             return None
 
     def fetchall(self):
         self._raise_if_closed()
 
-        return list(self.__iter__())
+        res = list(self.__iter__())
```
olavloite, line 317: Here also: Not directly relevant for this change, but this call can also cause an `Aborted` error that should initiate a transaction retry.

google/cloud/spanner_dbapi/cursor.py
```diff
@@ -335,7 +340,9 @@ def fetchmany(self, size=None):
         items = []
         for i in range(size):
             try:
-                items.append(tuple(self.__next__()))
+                res = next(self)
```
olavloite, line 343: Same as above: This can cause an `Aborted` error.

tests/spanner_dbapi/test_checksum.py
```diff
@@ -0,0 +1,44 @@
...
+
+from google.cloud.spanner_dbapi.checksum import (
+    _compare_checksums,
+    ResultsChecksum,
+)
+
+
+class Test_compare_checksums(unittest.TestCase):
+    def test_no_original_checksum(self):
+        self.assertIsNone(_compare_checksums(None, ResultsChecksum()))
```
olavloite, line 17: I would say that this should cause an error. Assume that the `_compare_checksums` function is called during a transaction retry. Then the following situations should cause the retry to be aborted because of differing results during the retry when compared to the initial attempt:
1. A query that did not return any results during the initial attempt does return results during the retry.
1. A query returns different results during the retry than the initial attempt.
1. A query returns an error during the retry, and did not return an error during the initial attempt (or vice-versa).

google/cloud/spanner_dbapi/checksum.py
```diff
@@ -0,0 +1,72 @@
...
+    :type original: :class:`~google.cloud.spanner_v1.transaction.ResultsChecksum`
+    :param original: results checksum of the original transaction.
+
+    :type retried: :class:`~google.cloud.spanner_v1.transaction.ResultsChecksum`
+    :param retried: results checksum of the retried transaction.
+
+    :raises: :exc:`RuntimeError` in case if checksums are not equal.
+    """
+    if original is not None:
+        if len(retried) == len(original) and retried != original:
```
olavloite, line 69: I think there's a misunderstanding regarding why we must keep track of `count` here. During a transaction retry we need to execute all statements again and check the results of the retried statements with the original statements. For queries there are two possibilities during a retry:
1. The results of a query have already been fully consumed by the client application (i.e. all rows have been fetched). In that case, during a retry, we must re-execute the statement and fetch all rows. Once we have done that, we can compare the checksums with each other. The `count` should also be equal for both result sets, but we do not necessarily need to check that, as if they were to be different, the checksums will also be different.
1. The results of a query have NOT yet been fully consumed by the client application (i.e. an `Aborted` error occurred while the client application was iterating over the result set). If that happens, we need to re-execute the statement and then advance the underlying iterator of the result set `count` steps and compare the checksums with each other. If the checksums are equal, the results that the client application has seen is equal to the results that we got during the retry, and the transaction may proceed. We also need to replace the underlying iterator of the result set that has been returned to the client application with the iterator that we got during the retry, as that iterator is now associated with the current transaction, while the old iterator is no longer valid. In the Java implementation, replacing the underlying iterator (`delegateResultSet`) [happens here](https://github.com/googleapis/java-spanner/blob/eddd6ad4e5093ee21290b85f15fa432d071bae59/google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ChecksumResultSet.java#L177).

In the Java implementation the step that is described

tests/spanner_dbapi/test_checksum.py
```diff
@@ -0,0 +1,44 @@
...
+
+        self.assertIsNone(_compare_checksums(original, retried))
+
+    def test_less_results(self):
+        original = ResultsChecksum()
+        original.consume_result(5)
+
+        retried = ResultsChecksum()
+
+        self.assertIsNone(_compare_checksums(original, retried))
```
olavloite, line 34: I think this is a little bit of a misunderstanding. Two result sets that have consumed different number of rows, should never have equal checksums.

google/cloud/spanner_dbapi/checksum.py
IlyaFaer: Yeah, that's something left from implementation in the original Spanner client. Changing to `Aborted`

google/cloud/spanner_dbapi/cursor.py
IlyaFaer: Yes, I'm pusing retry logic only today.

google/cloud/spanner_dbapi/cursor.py
```diff
@@ -305,14 +305,19 @@ def fetchone(self):
         self._raise_if_closed()
 
         try:
-            return next(self)
+            res = next(self)
+            self._checksum.consume_result(res)
```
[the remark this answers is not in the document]
IlyaFaer: Yes, it'll be a row or `StopIteration`

tests/spanner_dbapi/test_checksum.py
IlyaFaer: This logic is no more actually needed. It was done for the original Spanner client, but here it's pointless.

tests/spanner_dbapi/test_checksum.py
IlyaFaer: I've made it like this: checksum tracks the number of consumed results internally. When one is comparing two checksums, the first thing that happens is their length comparement. If they have equal length and different checksums, then the check will raise `Aborted` exception. In all other cases it'll do nothing and will return `None`.

google/cloud/spanner_dbapi/checksum.py
IlyaFaer: We're keeping counts inside checksum objects. When checksums are compared, they are first compared by length. If lengths are different, nothing will happen, as we consider this like: "the retried transaction is not yet came to the same point" or "the retried transaction have more results - it passed the fail point of the original transaction". But if the checksums have the same length and different checksums, it'll fail.

google/cloud/spanner_dbapi/connection.py
```diff
@@ -104,6 +108,25 @@ def _release_session(self):
...
+            If results checksum of the retried statement is
+            not equal to the checksum of the original one.
+        """
+        for statement in self._statements:
+            res_iter, retried_checksum = self.run_statement(
+                statement, retried=True
+            )
+            for res in res_iter:
+                retried_checksum.consume_result(res)
+                _compare_checksums(statement["checksum"], retried_checksum)
```
IlyaFaer, line 128: Here the lengths of the checksums will first be compared. If they are not equal, checksums themselves will not be compared.

IlyaFaer:
Pushed the second step of the implementation. Added actual retrying of aborted transaction. Requires to add more unit tests for `Cursor` - to check if `fetch*` methods are retrying transactions correctly.

tests/spanner_dbapi/test_connect.py
```diff
@@ -29,31 +29,20 @@ def test_connect(self):
...
+            )
 
-                self.assertIsInstance(connection, Connection)
-                client_info_mock.assert_called_once_with(USER_AGENT)
+            self.assertIsInstance(connection, Connection)
 
             client_mock.assert_called_once_with(
-                project=PROJECT,
-                credentials=CREDENTIALS,
-                client_info=CLIENT_INFO,
```
IlyaFaer, line 56: This test is actually broken, but nox doesn't see it and doesn't run it.

tests/spanner_dbapi/test_connection.py
```diff
@@ -11,7 +11,9 @@
 
 # import google.cloud.spanner_dbapi.exceptions as dbapi_exceptions
 
+from google.api_core.exceptions import Aborted
 from google.cloud.spanner_dbapi import Connection, InterfaceError
+from google.cloud.spanner_dbapi.checksum import ResultsChecksum
 from google.cloud.spanner_dbapi.connection import AUTOCOMMIT_MODE_WARNING
 from google.cloud.spanner_v1.database import Database
 from google.cloud.spanner_v1.instance import Instance
```
IlyaFaer, line 19: Why these tests are still here? They were copied into `unit` directory, so I suppose they should be erased from this directory?!

google/cloud/spanner_dbapi/cursor.py
```diff
@@ -207,9 +213,31 @@ def fetchone(self):
...
+            return res
         except StopIteration:
-            return None
+            return
+        except Aborted:
+            self.connection.retry_transaction()
+            return self.fetchone()
```
IlyaFaer, line 223: Should we add a retry limit here?

IlyaFaer, line 223: When we retried a transaction up to the failed row, we need to re-execute the failed `fetchone()`

IlyaFaer:
@olavloite, am I understood correctly, that `commit()` can also be aborted? And in this case we should retry the whole transaction?

olavloite:
> @olavloite, am I understood correctly, that `commit()` can also be aborted? And in this case we should retry the whole transaction?

@IlyaFaer Yes, that is correct.

google/cloud/spanner_dbapi/connection.py
```diff
@@ -198,9 +199,13 @@ def commit(self):
...
-            self._transaction.commit()
-            self._release_session()
-            self._statements = []
+            try:
+                self._transaction.commit()
+                self._release_session()
+                self._statements = []
+            except Aborted:
+                self.retry_transaction()
+                self.commit()
```
olavloite, line 208: The retry attempt itself + the `commit` call should all be placed inside a retry-loop, as the retry attempt itself could also abort, and should then also be retried.

tests/spanner_dbapi/test_connection.py
c24t: Good question. Looks like this (and `test_connect`) weren't moved in #532?

https://github.com/q-logic/python-spanner-django/blob/41abaebb6f2e0b1cf16704aa1e394acc5a47e68b/tests/spanner_dbapi/test_connection.py

The test files weren't exactly copied, #532 changed them and added some new tests. E.g. the [version on master now](https://github.com/googleapis/python-spanner-django/blob/927e17846398a111e3cd5bfe878d5497eded0609/tests/unit/spanner_dbapi/test_connection.py) doesn't include `test_transaction_autocommit_warnings`.

@mf2199 can you confirm that you meant to change/remove these tests before removing `tests/spanner_dbapi` in this PR?

tests/unit/spanner_dbapi/test_cursor.py
```diff
@@ -458,3 +467,121 @@ def test_get_table_column_schema(self):
...
+            "sql": "SELECT 1",
+            "params": [],
+            "param_types": {},
+            "checksum": cursor._checksum,
+        }
+        connection._statements.append(statement)
+
+        with mock.patch(
+            "google.cloud.spanner_dbapi.cursor.Cursor.__next__",
+            side_effect=(Aborted("Aborted"), None),
```
c24t, line 533: Will iterating through the result set raise `Aborted`? Looking at https://github.com/googleapis/python-spanner/blob/ca82c1f0ae005e01ea31b0d7664e4415820429b9/google/cloud/spanner_v1/gapic/spanner_client.py#L673-L675, I'd expect this to raise on the call to `execute_sql`.

google/cloud/spanner_dbapi/cursor.py
```diff
@@ -207,9 +214,31 @@ def fetchone(self):
...
+
+    def fetchall(self):
+        """Fetch all (remaining) rows of a query result, returning them as
+        a sequence of sequences.
+        """
+        self._raise_if_closed()
+
+        res = []
+        try:
+            for row in self.__iter__():
```
c24t, line 234:
```suggestion
            for row in self:
```

Nit: don't think you need the explicit `__iter__`.

google/cloud/spanner_dbapi/connection.py
```diff
@@ -190,6 +235,33 @@ def run_prior_DDL_statements(self):
...
+        transaction = self.transaction_checkout()
+        self._statements.append(statement)
+
+        return (
+            transaction.execute_sql(
+                statement["sql"],
+                statement["params"],
+                param_types=statement["param_types"],
+            ),
+            ResultsChecksum() if retried else statement["checksum"],
```
c24t, line 262: A [namedtuple](https://docs.python.org/3/library/collections.html#collections.namedtuple) would make this more readable (`statement.sql`, etc.).

google/cloud/spanner_dbapi/connection.py
```diff
@@ -160,8 +199,13 @@ def commit(self):
         if self._autocommit:
             warnings.warn(AUTOCOMMIT_MODE_WARNING, UserWarning, stacklevel=2)
         elif self._transaction:
-            self._transaction.commit()
-            self._release_session()
+            try:
+                self._transaction.commit()
+                self._release_session()
+                self._statements = []
```
c24t, line 205: Better to keep these out of the try block if you're only catching errors for the first line.

c24t (commented):
Besides cleaning up the tests, I wonder if this PR is catching `Aborted` errors in the right place. Since we're only calling `execute_sql` and not the streaming variant, I'd expect to see errors immediately instead of when we iterate over the results in the cursor. Am I missing something here?

google/cloud/spanner_dbapi/connection.py
IlyaFaer: Retry inside retry? Okay)

google/cloud/spanner_dbapi/connection.py
IlyaFaer: As I'm calling `self.commit()` on exception, these lines will be executed twice, if we'll move them outside the `try` block

google/cloud/spanner_dbapi/connection.py
olavloite: > Retry inside retry? Okay)

Yep. I haven't looked into the other retry implementation yet, but you might need to make this a little bit more generic. Any statement during a retry could of course also return an `Aborted` error, and any of those should also automatically restart the retry.

google/cloud/spanner_dbapi/cursor.py
IlyaFaer: That's true

tests/unit/spanner_dbapi/test_cursor.py
IlyaFaer: As I see, `Aborted` can be raised only while streaming:

Безымянный

By the way, `Aborted` never failed in tests in this branch, so I suppose we're on the right way.

google/cloud/spanner_dbapi/connection.py
IlyaFaer: NP, done

google/cloud/spanner_dbapi/cursor.py
```diff
@@ -207,9 +212,31 @@ def fetchone(self):
...
-            return next(self)
+            res = next(self)
+            self._checksum.consume_result(res)
+            return res
         except StopIteration:
-            return None
+            return
+        except Aborted:
+            self.connection.retry_transaction()
+            return self.fetchone()
```
olavloite, line 222: I think this will be a problem. Assuming that this is using the `ExecuteStreamingSql` RPC, then each `next()` call could potentially mean that a new RPC is executed. So for the sake of simplicity, assume in the example below that __each__ call to `next()` executes the `ExecuteStreamingSql` RPC.

Assume the following situation:
1. The table Singers contains the following singers (last names): Allison, Morrison, Pieterson
1. The application executes the query `SELECT LastName FROM Singers ORDER BY LastName` in transaction 1.
1. The client application calls `fetchone()` which returns 'Allison'.
1. Some other transaction executes `DELETE FROM Singers WHERE LastName='Pieterson'.
1. The first transaction is aborted by the backend. A retry is executed and the retry logic checks that the checksum of the retried result set is equal to the original attempt, which it is as the first record is still 'Allison'.
1. The client application calls `fetchone()` again. This should return 'Morrison', but as it needs to call `ExecuteStreamingSql` it will (probably) use the transaction id of the original transaction (unless that transaction id has somehow been replaced in the underlying iterator). If it does use the old transaction id, the RPC will fail with yet another `Aborted` error, and that will repeat itself until the transaction retry limit has been reached.

IlyaFaer: Seems to me we can just drop the `_transaction` property, so that `Connection` will initiate a new one on the next `execute()` call.

google/cloud/spanner_dbapi/exceptions.py
```diff
@@ -92,3 +92,13 @@ class NotSupportedError(DatabaseError):
     """
 
     pass
+
+
+class AbortedRetried(OperationalError):
```
olavloite, line 97: This should rather be called something like `RetryAborted`.

google/cloud/spanner_dbapi/connection.py
```diff
@@ -190,6 +265,33 @@ def run_prior_DDL_statements(self):
...
+        :type statement: :class:`dict`
+        :param statement: SQL statement to execute.
+
+        :rtype: :class:`google.cloud.spanner_v1.streamed.StreamedResultSet`,
+                :class:`google.cloud.spanner_dbapi.checksum.ResultsChecksum`
+        :returns: Streamed result set of the statement and a
+                  checksum of this statement results.
+        """
+        transaction = self.transaction_checkout()
+        self._statements.append(statement)
```
olavloite, line 284: I think we should only add the statement here if `retried=False`. Otherwise, if the transaction is aborted and retried multiple times, the list of statements that are executed on the transaction is duplicated.

google/cloud/spanner_dbapi/connection.py
```diff
@@ -190,6 +265,33 @@ def run_prior_DDL_statements(self):
...
+
+        return (
+            transaction.execute_sql(
```
olavloite, line 287: Is this method using the `executeStreamingSql` RPC? Or is it using the `executeSql` RPC?

olavloite (commented):
Thanks for the updates. I think this is coming very to close to what we need, but I have a couple of small questions.

google/cloud/spanner_dbapi/exceptions.py
IlyaFaer: Done

google/cloud/spanner_dbapi/connection.py
IlyaFaer: Oh, yeah, I'm only clearing statements on `commit()` and `rollback()` for now. Good catch!

google/cloud/spanner_dbapi/connection.py
IlyaFaer: This method is inherited from `_SnapshotBase` class, here it is:
https://github.com/googleapis/python-spanner/blob/783fffb6683119db0dde89bff66112e451881c19/google/cloud/spanner_v1/snapshot.py#L178-L189

So, it's `ExecuteStreamingSql`.

c24t (commented):
No more substantial comments from me, but note that this whole PR will have to move to python-spanner now that https://github.com/googleapis/python-spanner/pull/160 is in.

@olavloite, does this address all your comments?

olavloite:
> No more substantial comments from me, but note that this whole PR will have to move to python-spanner now that [googleapis/python-spanner#160](https://github.com/googleapis/python-spanner/pull/160) is in.
> 
> @olavloite, does this address all your comments?

Yes, my comments have been addressed. I do think that it's important that we get the (integration) tests running soon, as there could still be corner cases that we haven't thought of yet that only occur incidentally.

google/cloud/spanner_dbapi/cursor.py
olavloite: @IlyaFaer @c24t 

Sorry for reopening this, and this comment should not be considered blocking for merging this PR, but I think we need to look into this once more. Only dropping the `_transaction` property will in this case not be enough for the following reason:
1. When `executeSql` is called, a streaming iterator is returned to the application.
1. That streaming iterator is linked with the transaction that was active at that moment, and a reference to that transaction is also [held in the iterator](https://github.com/googleapis/python-spanner/blob/783fffb6683119db0dde89bff66112e451881c19/google/cloud/spanner_v1/snapshot.py#L257).
1. If a transaction is aborted and the client application has consumed only parts of a streaming iterator, that iterator is no longer valid (at least: it will also throw an exception if it needs to receive more data from the server).

The JDBC driver client solves the above problem by [wrapping all streaming iterators](https://github.com/googleapis/java-spanner/blob/3fa7910c8e5089cff1c9ed645f160a9e0ddfc351/google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ReplaceableForwardingResultSet.java#L44) before returning these to the client application. That makes it possible for the JDBC driver to [replace the underlying streaming iterator](https://github.com/googleapis/java-spanner/blob/3fa7910c8e5089cff1c9ed645f160a9e0ddfc351/google-cloud-spanner/src/main/java/com/google/cloud/spanner/connection/ChecksumResultSet.java#L177) with a new one when a transaction has been aborted and successfully retried.

We should add that to the Python DBApi as well, but we could do that in a separate PR to prevent this PR from becoming even bigger than it already is.

IlyaFaer: @c24t, @olavloite, hm-m. I think we're protected from errors here, because our connection API doesn't actually give streaming result objects to a user.

Here is where we're getting a streaming iterator:
https://github.com/googleapis/python-spanner-django/blob/196c44949370fb818e610b21c3b00344fdc3d03a/google/cloud/spanner_dbapi/cursor.py#L167-L170

So, iterator is held in the protected property `_itr`, and users will be streaming it with `Cursor.fetch*()` methods, without actual access to the iterator itself:
https://github.com/googleapis/python-spanner-django/blob/196c44949370fb818e610b21c3b00344fdc3d03a/google/cloud/spanner_dbapi/cursor.py#L204-L212

Where `next(self)` is calling `next(self._itr)` here:
https://github.com/googleapis/python-spanner-django/blob/196c44949370fb818e610b21c3b00344fdc3d03a/google/cloud/spanner_dbapi/cursor.py#L293-L296

Thus, if a transaction failed, the connection will drop the transaction, checkout a new one, re-run all the statements, each of which will replace `_itr` with new streamed iterator. So, all the iterators are processed internally, and will be replaced on a retry, as I see.

c24t (approved):

c24t:
This PR won't be merged here, see https://github.com/googleapis/python-spanner/pull/168.

IlyaFaer:
Merged into the original Spanner client repo

PR #545, opened 2020-10-25
fix: missing method implementation and potential bug in the `DatabaseWrapper` class

mf2199:
Change list:

1. Added implementation of `DatabaseWrapper.init_connection_state()` method;
2. Potential bug fix in the `DatabaseWrappper.is_usable()` condition check;
3. Miscellaneous refactoring and reformatting.


IlyaFaer (approved):

django_spanner/base.py
```diff
@@ -124,7 +127,9 @@ def get_new_connection(self, conn_params):
         return Database.connect(**conn_params)
 
     def init_connection_state(self):
-        pass
+        self.connection.close()
```
c24t, line 130: Won't this create the new connection and then immediately close it? https://github.com/django/django/blob/8093aaa8ff9dd7386a069c6eb49fcc1c5980c033/django/db/backends/base/base.py#L195-L197

c24t (commented):
What's the potential bug this fixes? Do you expect `init_connection_state` to get called after initializing the wrapper?

Can you add a test that fails on master but passes with this change?

django_spanner/base.py
mf2199: There is no `connection` property hence no active getter to be called, so `self.connection.close()` directly invokes the `Connection.close()` method that simply rolls back the outstanding transaction, if any, and sets the `is_closed` flag to `True` without destroying the connection object itself. Then we invoke the `__init__()` method that initializes the connection state.

Answering the question, no, this will not create a new connection (in its current implementation) and immediately close it, but rather reset the existing connection to its initial state.

version.py
```diff
@@ -4,4 +4,4 @@
 # license that can be found in the LICENSE file or at
 # https://developers.google.com/open-source/licenses/bsd
 
-__version__ = "2.2.0a1"
+__version__ = "3.1.0a1"
```
mf2199, line 7: This is necessary to avoid compatibility [check](https://github.com/googleapis/python-spanner-django/blob/196c44949370fb818e610b21c3b00344fdc3d03a/django_spanner/utils.py#L6) failure.

tests/unit/django_spanner/test_base.py
```diff
@@ -0,0 +1,108 @@
...
+
+import sys
+import unittest
+
+from mock_import import mock_import
+from unittest import mock
+
+
+@mock_import()
+@unittest.skipIf(sys.version_info < (3, 6), reason="Skipping Python 3.5")
```
c24t, line 15: I think we can safely drop 3.5 in tests since we already decided not to support it elsewhere.

django_spanner/base.py
```diff
@@ -92,7 +95,7 @@ class DatabaseWrapper(BaseDatabaseWrapper):
         "iendswith": "",
     }
 
-    Database = Database
+    Database = spanner_dbapi
```
c24t, line 98: FWIW this odd "import x as Database; Database = Database" pattern seems to come from sqlite (https://github.com/django/django/blob/e893c0ad8b0b5b0a1e5be3345c287044868effc4/django/db/backends/sqlite3/base.py#L156), and is copied in all the other built in backends.

c24t (approved):
Looks fine to me without requiring django 3.x.

Issue #518, opened 2020-09-22
Cursor.executemany() hiding results up to the last one


IlyaFaer:
I think we have a bug in `Cursor.executemany()` method. Given the code it runs `Cursor.execute()` in a cycle, while `execute()` is actually replacing result of every previous call with a result from the last one.

So, if we, for example, will call `executemany()` with several reads, we'll get only results of the last one:

![Безымянный](https://user-images.githubusercontent.com/26030071/93857979-20a9a600-fcc4-11ea-81f3-5eb4069592e4.png)

But in fact both reads, with `SingerId=12` and with `SingerId=15`, were executed and we got results.

@c24t, we already have a proposition of how to deal with `executemany()` in the design doc, PTAL at `StreamedManyResultSets` section.

skuruppu:
@IlyaFaer this was turning up on SLO dashboards so I put a priority on it. Please feel free to change the priority to what you want.

IlyaFaer:
@c24t, are we okay to deal with it by adding the class `StreamedManyResultSets`, described in the design doc? I think the transactions retry is closing to the final state, so I'll do this one as well.

c24t:
From @IlyaFaer's doc:

New `StreamedManyResultSets` class:
A wrap-around iterator, which iterates through several `StreamedResultSet()` as if they are all merged into a single iterator. This class is intended to be used by `Cursor.executemany()` method. After executing a query the method will save the resulting `StreamedResultSet()` object or `int` in case of an UPDATE operation into `StreamedManyResultSets`, before running the next query (which results will replace the results of the previous operation given the current implementation).

Thus, `Cursor.executemany()` will pump up all the resulting iterators or integers into a single `StreamedManyResultSets` iterator. While iterating through it with `fetchone()`, `fetchmany()` and `fetchall()`, users will be streaming through the results of every executed operation in the same order the operations were executed.

IlyaFaer:
This one was fixed in the [python-spanner PR](https://github.com/googleapis/python-spanner/pull/181). Closing

PR #551, opened 2020-11-03
fix: executemany() should return every executed query result

IlyaFaer:
Fixes: #518


google/cloud/spanner_dbapi/cursor.py
```diff
@@ -198,8 +202,12 @@ def executemany(self, operation, seq_of_params):
         """
         self._raise_if_closed()
 
+        self._many_result_set = StreamedManyResultSets()
```
c24t, line 205: I don't know that you need a separate class for this, you could [chain the `self._itr` calls](https://docs.python.org/3/library/itertools.html#itertools.chain.from_iterable) instead to do the same thing.

It may still be useful to have a class to e.g. get the length of the combined result set (although it looks like `StreamedResultSet` itself doesn't have a length), but I don't think you need it for what it's doing now.

c24t (commented):
Comments about setting the result set and row count.

Chaining the result sets from each `execute` call looks sane to me, but FWIW [PEP 249 makes it sound like this isn't the typical use](https://www.python.org/dev/peps/pep-0249/#executemany): 

> Use of this method for an operation which produces one or more result sets constitutes undefined behavior, and the implementation is permitted (but not required) to raise an exception when it detects that a result set has been created by an invocation of the operation.

`executemany` shouldn't allow DDL right? I ask because it looks like `execute` queues up DDL statements without executing them and only calls `update_ddl` on the next non-DDL call. Also [sqlite disallows this](https://github.com/python/cpython/blob/master/Modules/_sqlite/cursor.c#L582-L586).

tests/system/test_system.py
```diff
@@ -289,6 +289,46 @@ def test_rollback_on_connection_closing(self):
...
+    def test_execute_many(self):
+        # connect to the test database
+        conn = Connection(Config.INSTANCE, self._db)
+        cursor = conn.cursor()
+
+        cursor.execute(
+            """
+INSERT INTO contacts (contact_id, first_name, last_name, email)
+VALUES (1, 'first-name', 'last-name', 'test.email@domen.ru'),
+       (2, 'first-name2', 'last-name2', 'test.email2@domen.ru')
```
c24t, line 301:
```suggestion
VALUES (1, 'first-name', 'last-name', 'test.email@example.com'),
       (2, 'first-name2', 'last-name2', 'test.email2@example.com')
```

Not a big deal, but we probably want to avoid including non-google, non-example domains anywhere, even in test code.

google/cloud/spanner_dbapi/cursor.py
```diff
@@ -198,8 +202,12 @@ def executemany(self, operation, seq_of_params):
...
         for params in seq_of_params:
             self.execute(operation, params)
+            self._many_result_set.add_iter(self._itr)
```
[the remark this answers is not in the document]
IlyaFaer: Hm-m, looking at this once again, it seems like we don't actually need new properties.

IlyaFaer:
> executemany shouldn't allow DDL right? I ask because it looks like execute queues up DDL statements without executing them and only calls update_ddl on the next non-DDL call

@c24t, okay, fixed.

c24t:
Closed in favor of https://github.com/googleapis/python-spanner/pull/181.

PR #552, opened 2020-11-05
feat: autocommit sample

AlisskaPie:
This is one of the samples that demonstrates the difference between autocommit and !autocommit modes


AlisskaPie:
I got the information that this repository will be moved to python-spanner within a week or so. And there we could check test's passing, because kokoro failed of other code fails. 
Is it ok to have this sample?

samples/samples/test_autocommit.py
```diff
@@ -0,0 +1,53 @@
...
+@pytest.fixture(scope="module")
+def database(spanner_instance):
+    """Creates a temporary database that is removed after testing."""
+    db = spanner_instance.database(DATABASE_ID)
+    db.create()
+    yield db
+    db.drop()
+
+
+def test_enable_autocommit_mode(capsys):
```
c24t, line 49:
```suggestion
def test_enable_autocommit_mode(capsys, database):
```

Looks like `database` is unused? I get this error without this change:

```
...
>           raise ValueError("instance '%s' does not exist." % instance_id)
E           ValueError: instance 'test-instance-86c4b1fe62' does not exist.

google/cloud/spanner_dbapi/connection.py:262: ValueError
```

samples/samples/test_autocommit.py
```diff
@@ -0,0 +1,53 @@
...
+# license that can be found in the LICENSE file or at
+# https://developers.google.com/open-source/licenses/bsd
+
+import uuid
+
+from google.api_core.exceptions import DeadlineExceeded
+from google.cloud import spanner
+import pytest
+
+import autocommit
```
c24t, line 13:
```suggestion
from . import autocommit
```

AFAIK it's never a good idea to use old-style relative imports, but maybe this is the recommended style for standalone docs?

Note that you'll need an `__init__.py` file in this package for this suggestion to work.

samples/samples/autocommit.py
```diff
@@ -0,0 +1,69 @@
...
+
+    print(
+        u"SingerId: {}, AlbumId: {}, AlbumTitle: {}".format(*cursor.fetchone())
+    )
+
+    connection.close()
+    # [END enable_autocommit_mode]
+
+
+if __name__ == "__main__":
```
c24t, line 48: Should this file be executable?

samples/samples/autocommit.py
```diff
@@ -0,0 +1,69 @@
+# Copyright 2020 Google LLC
+#
+# Use of this source code is governed by a BSD-style
+# license that can be found in the LICENSE file or at
+# https://developers.google.com/open-source/licenses/bsd
+
+import argparse
+
+from google.cloud.spanner_dbapi import connect
```
c24t, line 9: As a heads up: this import will fail unless you put `spanner_dbapi` on the path yourself (calling `import pkg_resources` first will do it) until we fix the DBAPI packaging issues.

samples/samples/autocommit.py
```diff
@@ -0,0 +1,69 @@
...
+
+
+def enable_autocommit_mode(instance_id, database_id):
+    """Enables autocommit mode."""
+    # [START enable_autocommit_mode]
```
c24t, line 14: I see this matches spanner-python, the point is to use these samples in embedded docs?

c24t (commented):
A few comments, but the sample looks good and the test works after making a small change.

I see this sample describes _how_ to enable autocommit, but doesn't show the difference between auto and non-auto behavior. Ideally this sample would show multiple queries each getting a separate txn in autocommit mode and getting combined in a single txn in non-auto mode. And would show manual `commit` calls.

Not a problem with this sample, but a problem with the library: we should respect [django's autocommit setting](https://docs.djangoproject.com/en/3.1/topics/db/transactions/#managing-autocommit). When DBAPI  is moved out of this library I wouldn't expect users to set `connection.autocommit` themselves.

samples/samples/autocommit.py
c24t: Got it, these are docs snippets.

AlisskaPie (commented):
> I see this sample describes _how_ to enable autocommit, but doesn't show the difference between auto and non-auto behavior. Ideally this sample would show multiple queries each getting a separate txn in autocommit mode and getting combined in a single txn in non-auto mode. And would show manual `commit` calls.
> 
I'm sorry, I wasn't correct in a comment and the sample doesn't show the difference between those two modes. I wanted to say that it shows only one of the modes. And for other mode there will be another sample. So users will see that way the difference themselves. 
> Not a problem with this sample, but a problem with the library: we should respect [django's autocommit setting](https://docs.djangoproject.com/en/3.1/topics/db/transactions/#managing-autocommit). When DBAPI is moved out of this library I wouldn't expect users to set `connection.autocommit` themselves.

Just to clarify: I need to use [`set_autocommit`](https://docs.djangoproject.com/en/3.1/topics/db/transactions/#django.db.transaction.set_autocommit) function instead?

c24t:
> And for other mode there will be another sample. So users will see that way the difference themselves.

That sounds fine, let's see how these samples appear in the cloud docs.

> Just to clarify: I need to use [`set_autocommit`](https://docs.djangoproject.com/en/3.1/topics/db/transactions/#django.db.transaction.set_autocommit) function instead?

Ideally yes, this would just follow django's autocommit setting and the user wouldn't set this at all. But now since https://github.com/googleapis/python-spanner/pull/160 moved the DBAPI package into python-spanner we might need to move these samples over too, and write different django-specific docs for this spanner-django.

samples/samples/autocommit.py
c24t: I just meant to ask whether these files should have the `x` bit set so you can run them as `./samples/samples/autocommit.py` instead of `python ...`, but I see the python-spanner samples don't.

c24t (approved):
LGTM, but let's check whether this should move into python-spanner with https://github.com/googleapis/python-spanner/pull/160 before merging here.

tmatsuo:
> There is a possible violation for not having product prefix.

snippet-bot thinks it's violating a region tag rule where we should have product prefix. If you change the region tag to `spanner_enable_autocommit_mode`, the bot will be happy.

samples/samples/autocommit.py
```diff
@@ -0,0 +1,70 @@
...
+
+import argparse
+
+from google.cloud import spanner_dbapi
+from google.cloud.spanner_dbapi import connect
+
+
+def enable_autocommit_mode(instance_id, database_id):
+    """Enables autocommit mode."""
+    # [START enable_autocommit_mode]
```
tmatsuo, line 15: Can you change it to `spanner_enable_autocommit_mode`?

samples/samples/autocommit.py
```diff
@@ -0,0 +1,70 @@
...
+    )
+
+    cursor.execute("""SELECT * FROM Singers WHERE SingerId = 13""")
+
+    print(
+        u"SingerId: {}, AlbumId: {}, AlbumTitle: {}".format(*cursor.fetchone())
+    )
+
+    connection.close()
+    # [END enable_autocommit_mode]
```
tmatsuo, line 46: ditto

tmatsuo (changes requested):

AlisskaPie:
> > There is a possible violation for not having product prefix.
> 
> snippet-bot thinks it's violating a region tag rule where we should have product prefix. If you change the region tag to `spanner_enable_autocommit_mode`, the bot will be happy.

Thank you! Bot is happy now :tada:

tmatsuo (approved):

dinagraves:
The "Create Database" part of your tests is failing b/c the database already exists. 

From the logs: 
```
E           	status = StatusCode.ALREADY_EXISTS
E           	details = "Database already exists: projects/precise-truck-742/instances/google-cloud-python-systest/databases/db-api-transactions-management"
```

samples/samples/autocommit.py
```diff
@@ -0,0 +1,70 @@
...
+    subparsers = parser.add_subparsers(dest="command")
+    subparsers.add_parser(
+        "enable_autocommit_mode", help=enable_autocommit_mode.__doc__
+    )
+    args = parser.parse_args()
+    if args.command == "enable_autocommit_mode":
+        enable_autocommit_mode(args.instance_id, args.database_id)
+    else:
+        print(
+            "Command {} did not match expected commands.".format(args.command)
```
dinagraves, line 69: nit: f-strings are preferable.  More readable.

samples/samples/test_autocommit.py
```diff
@@ -0,0 +1,56 @@
...
+from google.api_core.exceptions import DeadlineExceeded
+from google.cloud import spanner
+import pytest
+
+from . import autocommit
+
+
+def unique_instance_id():
+    """Creates a unique id for the database."""
+    return "test-instance-{}".format(uuid.uuid4().hex[:10])
```
dinagraves, line 18: nit 
```suggestion
    return f"test-instance-{uuid.uuid4().hex[:10]}"
```

samples/samples/test_autocommit.py
```diff
@@ -0,0 +1,56 @@
...
+
+
+def unique_database_id():
+    """Creates a unique id for the database."""
+    return "test-db-{}".format(uuid.uuid4().hex[:10])
```
dinagraves, line 23: nit 

```suggestion
    return f"test-db-{uuid.uuid4().hex[:10]}"
```

samples/samples/test_autocommit.py
```diff
@@ -0,0 +1,56 @@
...
+
+
+INSTANCE_ID = unique_instance_id()
+DATABASE_ID = unique_database_id()
+
+
+@pytest.fixture(scope="module")
+def spanner_instance():
+    spanner_client = spanner.Client()
+    config_name = "{}/instanceConfigs/regional-us-central1".format(
```
dinagraves, line 33: nit: f-string

dinagraves (approved):
Now that the tests are passing, I only have formatting nits.

samples/samples/autocommit.py
AlisskaPie: Done

samples/samples/test_autocommit.py
AlisskaPie: Done

samples/samples/test_autocommit.py
AlisskaPie: Done

samples/samples/test_autocommit.py
AlisskaPie: Done

AlisskaPie (commented):
Thanks for your help!

c24t:
@AlisskaPie can we close this in favor of https://github.com/googleapis/python-spanner/pull/172?

AlisskaPie:
> @AlisskaPie can we close this in favor of [googleapis/python-spanner#172](https://github.com/googleapis/python-spanner/pull/172)?

Yes, sure! Sorry I forgot about it. 
Thanks!

PR #554, opened 2020-11-19
refactor!: erase dbapi directory and all the related tests

IlyaFaer:
BREAKING CHANGE: all the connection API related code was moved into the [original Cloud Spanner client repository](https://github.com/googleapis/python-spanner/tree/master/google/cloud/spanner_dbapi). With this PR the whole `dbapi` directory is erased from this repository.

The dependency on `google-cloud-spanner` of this API is bumped to 2.0.0.


IlyaFaer:
@c24t , system tests on emulator has failed, as there are no system tests at all now. Should we disable the workflow?

c24t (approved):
> @c24t, system tests on emulator has failed, as there are no system tests at all now. Should we disable the workflow?

I think so, and the deletion LGTM.

IlyaFaer:
@c24t, unit tests are not able to install `google-cloud-spanner` >= 2.0.0 somewhy, though I see it is released.

c24t:
> @c24t, unit tests are not able to install `google-cloud-spanner` >= 2.0.0 somewhy, though I see it is released.

Looks like this is because they dropped support for 3.5, fix in #557.

IlyaFaer:
@c24t, as we don't actually have unit tests now, we can just comment out these nox sessions until better times.

c24t:
@IlyaFaer to follow up on this PR with another change that re-enables django system tests against the emulator.

PR #559, opened 2020-11-30
test: run Django tests against Spanner emulator

IlyaFaer:


.github/workflows/django_tests_against_emulator.yml
```diff
@@ -0,0 +1,33 @@
...
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+      - name: Setup Python
+        uses: actions/setup-python@v2
+        with:
+          python-version: 3.8
+      - name: Run system tests
+        run: sh django_test_suite.sh
```
IlyaFaer, line 26: Running `django_test_suite.sh` file to prepare for testing and run tests

create_test_instance.py
```diff
@@ -0,0 +1,26 @@
...
+from google.cloud.spanner_v1 import Client
+
+
+client = Client(
+    project=os.getenv("GOOGLE_CLOUD_PROJECT", "emulator-test-project")
+)
+
+instance = client.instance("google-cloud-django-backend-tests")
+created_op = instance.create()
+created_op.result(30)  # block until completion
```
IlyaFaer, line 26: This script will be runned before tests

django_spanner/base.py
```diff
@@ -113,7 +117,7 @@ def _nodb_connection(self):
 
     def get_connection_params(self):
         return {
-            "project": self.settings_dict["PROJECT"],
+            "project": os.environ["GOOGLE_CLOUD_PROJECT"],
```
IlyaFaer, line 120: Few fixes to run the tests right on the emulator

django_test_suite.sh
```diff
@@ -42,21 +63,11 @@ PASSWORD_HASHERS = [
...
-        SPANNER_INSTANCE=$INSTANCE python3 .kokoro/ensure_instance_exists.py
-    fi
-}
-
 run_django_tests() {
     cd $TESTS_DIR/django/tests
     create_settings
     echo -e "\033[32mRunning Django tests: $TEST_APPS\033[00m"
-    python3 runtests.py $TEST_APPS --verbosity=2 --noinput --settings $SETTINGS_FILE
+    python3 runtests.py $TEST_APPS --verbosity=3 --noinput --settings $SETTINGS_FILE
```
IlyaFaer, line 70: I've moved a part of the `build.sh` to this file to reduce the number of scripts calling each other

IlyaFaer (commented):
Now we can drop `parallelize_tests.go`, its binary version, `build.sh` and several environment variables. I'm still keeping them in here, until all the things fixed.

As for the transaction commited error, I've pushed a PR with fix in the original Spanner repo:
https://github.com/googleapis/python-spanner/pull/177

c24t (commented):
A few comments, but looks like a big improvement over running the django tests against non-emulated spanner.

I'd like to see the tests passing before we merge this, but it looks like that depends on https://github.com/googleapis/python-spanner/pull/177. Did you check that the tests pass using your own fork of python-spanner with this change applied?

.github/workflows/django_tests_against_emulator.yml
```diff
@@ -0,0 +1,33 @@
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+name: Run Django backend tests against emulator
+jobs:
+  system-tests:
```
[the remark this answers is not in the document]
IlyaFaer: Okay

.github/workflows/django_tests_against_emulator.yml
```diff
@@ -0,0 +1,33 @@
...
+          - 9020:9020
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+      - name: Setup Python
+        uses: actions/setup-python@v2
+        with:
+          python-version: 3.8
+      - name: Run system tests
```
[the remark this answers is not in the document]
IlyaFaer: This is one is needed for detailed view in check logs. So, I'd keep it different.

.github/workflows/django_tests_against_emulator.yml
```diff
@@ -0,0 +1,33 @@
...
+        run: sh django_test_suite.sh
+        env:
+          SPANNER_EMULATOR_HOST: localhost:9010
+          GOOGLE_CLOUD_PROJECT: emulator-test-project
+          GOOGLE_CLOUD_TESTS_CREATE_SPANNER_INSTANCE: true
+          DJANGO_WORKER_INDEX: 0
+          DJANGO_WORKER_COUNT: 1
```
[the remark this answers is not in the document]
IlyaFaer: Yeah, we don't need them on emulator, but kokoro checks are still using them. I've decided to keep them until we'll completely clear all the old checks.

create_test_instance.py
IlyaFaer: Instance creation is a long operation, so we're setting 30 secs timeout. Instance clearing is made inside our Django code.

django_spanner/base.py
IlyaFaer: Hm-m, this requires a careful look. Some parts of piping client to emulator code uses env variables. I suppose setting both evn vars and the same Django settings is not something we should do.

django_test_suite.sh
```diff
@@ -5,7 +5,26 @@
 # license that can be found in the LICENSE file.
 
 # exit when any command fails
-set -e
+set -x pipefail
```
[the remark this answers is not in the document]
IlyaFaer: Sure, I'm just using it for debugging yet

django_test_suite.sh
IlyaFaer: For debugging only. There was no traceback at all, it was just showing `no Google credentials`, but in fact there was another error. So, yeah, after https://github.com/googleapis/python-spanner/pull/177 fix I'm gonna cleanup this PR once again.

IlyaFaer:
@c24t, now we have 9 Django tests failing because of several problems. `DELETE must have WHERE clause` error isn't dependent on adding dummy WHERE clause (I've already tried to revert it), so it requires some digging. Working on it.

django_test_suite.sh
```diff
@@ -5,7 +5,26 @@
...
+# Disable buffering, so that the logs stream through.
+export PYTHONUNBUFFERED=1
+
+# Export essential environment variables for Django tests.
+export RUNNING_SPANNER_BACKEND_TESTS=1
+
+pip3 install .
+# Create a unique DJANGO_TESTS_DIR per worker to avoid
+# any clashes with configured tests by other workers.
+export DJANGO_TESTS_DIR="django_tests_$DJANGO_WORKER_INDEX"
```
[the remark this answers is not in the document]
IlyaFaer: Yes, that's in use for now

IlyaFaer:
@c24t, I've got a green light on Django tests. There are still some things to refactor and improve, but I suppose it's not something to block the release.

IlyaFaer:
@c24t, okay, I've found the place where the `basic` tests were set to be executed, and fixed it.

But I'm little bit stuck with hitting 950 query args quota. Not sure why this error showed yet, everything seems to be set correctly: skips, dependencies, tests things like `max_query_params`.

IlyaFaer:
@c24t, so, the Django tests on emulator are canceled, because of a timeout hit. Seems to me we still need parallelisation, as slow execution problem has not gone with the emulator.

IlyaFaer:
@c24t, okay, 8 threads seems to be working, but there is a lot of `Primary key already exists` error. Aren't emulators conflicting with each other? Maybe we could use separate workflows?

IlyaFaer:
@c24t, I think I see the problem in the last 17 tests. All of them are walking through `_restart_on_unavailable()` Spanner method.

Issue #438, opened 2020-04-22
performance testing harness


odeke-em:
I've been requested that we should set up a performance test harness. Very tricky one because we can't directly measure against the client libraries. An apples to apples comparison would perhaps
be using spanner_dbapi, but django-spanner itself is tricky to set up.

odeke-em:
A while back, I create a benchmarking repository for spanner_dbapi vs the Python client library, at https://github.com/orijtech/spanner-benchmarking
which uses OpenCensus and Prometheus to visualize results e.g.

![Screen Shot 2020-04-22 at 2 50 17 PM](https://user-images.githubusercontent.com/4898263/80037556-a53f9080-84a8-11ea-988d-225fc58bc134.png)

odeke-em:
I've spun up this code https://gist.github.com/odeke-em/929d308f1fb80a1661d2e3ccb6a7b58b which produces ineresting results when comparing benchmarks https://gist.github.com/odeke-em/929d308f1fb80a1661d2e3ccb6a7b58b#file-results-md
```shell
name                old time/op  new time/op  delta
ReadOnly            93.7ms ±18%  87.6ms ±11%   -6.50%  (p=0.007 n=20+18)
ReadWrite           87.5ms ±13%  91.9ms ±39%     ~     (p=0.153 n=96+83)
CreateDestroyTable   47.3s ±20%   17.7s ±47%  -62.49%  (p=0.002 n=6+6)
```

prakhargurunani:
@odeke-em Can you tell me more about how to return results (display them in terminal or store them somewhere)

asthamohta:
#675 We have also done benchmark testing here where Spanner vs Django performance are tested

PR #562, opened 2020-12-01
[WIP] Spanner benchmarking

FirePing32:
fixes: #438


c24t:
Looks like `benchmarking.py` is taken from @odeke-em's gist here: https://gist.github.com/odeke-em/929d308f1fb80a1661d2e3ccb6a7b58b.

As @odeke-em mentions in https://github.com/googleapis/python-spanner-django/issues/438#issuecomment-618059406, what we're really interested in here is the extra latency that spanner-django adds on top of python-spanner. The python-spanner repo includes some [YCSB benchmarks](https://github.com/googleapis/python-spanner/blob/cf87cdf6a09e4fe3ac3e4009c3919c6223fb8ac9/benchmark/ycsb.py) (added in https://github.com/googleapis/google-cloud-python/pull/4539), but it's not clear that they're measuring the right thing, or that we can get a meaningful comparison in django.

Ideally we'd run these benchmarks as a periodic CI task and flag regressions automatically instead of checking the results into the repo.

c24t:
I suspect we'll want the benchmarking script to run against the emulator since we're only interested in the performance of the ORM library here, and not the spanner service itself.

c24t:
Closing as inactive, @FirePing32 feel free to reopen if you change this to address the comments above (run benchmarks in CI, against emulator, etc.).

PR #564, opened 2020-12-03
docs: Docstrings for `django_spanner`

mf2199:


mf2199:
@googlebot I consent.

tina80lvl:
@googlebot I consent.

django_spanner/client.py
```diff
@@ -9,5 +9,11 @@
 
 
 class DatabaseClient(BaseDatabaseClient):
+    """Wraps the Django base class.
+
+    TODO: Consider actual implementation of `runshell` method.
```
c24t, line 14: Please make this an issue instead of in the docstring.

django_spanner/expressions.py
```diff
@@ -8,8 +8,16 @@
...
 
 def order_by(self, compiler, connection, **extra_context):
-    # In Django 3.1, this can be replaced with
-    # DatabaseFeatures.supports_order_by_nulls_modifier = False.
+    """Order expressions in the SQL query and generate a new query using
+    Spanner-specific templates.
+
+    TODO: In Django 3.1, this can be replaced with
+     DatabaseFeatures.supports_order_by_nulls_modifier = False.
+     Also, consider making this function a part of a class.
```
c24t, line 16: This too, it can be a regular code comment but shouldn't be in the docstring.

django_spanner/client.py
mf2199: Moved into a regular code comment for now.

django_spanner/expressions.py
mf2199: Same here.

django_spanner/schema.py
```diff
@@ -123,6 +130,13 @@ def create_model(self, model):
                 self.create_model(field.remote_field.through)
 
     def delete_model(self, model):
+        """
+        Drop the model’s table in the database along with any unique constraints
```
c24t, line 134: Looks like this is a [unicode RIGHT SINGLE QUOTATION MARK](https://www.fileformat.info/info/unicode/char/2019/index.htm), did you write these docstrings in a rich text editor first?

c24t (commented):
The PR looks like a clear improvement (aside from the odd unicode chars), any reason it's still in draft?

django_spanner/schema.py
tina80lvl: Fixed it.

tina80lvl:
> The PR looks like a clear improvement (aside from the odd unicode chars), any reason it's still in draft?

I guess there is no such reason for now, I've fixed the rest changes.

django_spanner/client.py
```diff
@@ -9,5 +9,16 @@
 
 
 class DatabaseClient(BaseDatabaseClient):
+    """Wrap the Django base class."""
+
     def runshell(self, parameters):
+        """Override the base class method.
+
+        TODO: Consider actual implementation of this method.
```
c24t: It's not clear that we could implement this since cloud spanner doesn't have a "shell" executable. We might want to consider https://github.com/cloudspannerecosystem/spanner-cli or a similar project for this in the future.

django_spanner/operations.py
```diff
@@ -102,12 +190,35 @@ def adapt_decimalfield_value(
...
         if value is None:
             return None
         return float(value)
 
     def adapt_timefield_value(self, value):
+        """
+        Transform a time value to an object compatible with what is expected
+        by the backend driver for time columns.
+
+        :type value: #TODO
```
c24t, line 215: Looks like there are still a lot of TODOs in the docstrings in this file.

c24t (commented):
Please fix the TODOs.

django_spanner/client.py
c24t: I removed the docstring here, just raising `NotSupportedError` is fine since spanner doesn't really give us a way to support this.

django_spanner/operations.py
```diff
@@ -133,13 +251,32 @@ def get_db_converters(self, expression):
             converters.append(self.convert_uuidfield_value)
         return converters
 
+    # TODO: here and below, remove unused expression and connection parameters
+    # or include them to the code.
```
c24t, line 255: @mf2199 I moved this comment in https://github.com/googleapis/python-spanner-django/pull/564/commits/d36b9abd628dc2fd4a2d4535583d8dfc4ada3b73 since the preceding change had this comment in each of the converter docstrings:

> #TODO remove unused expression and connection parameters or include them to the code.

But converters are part of the django db API, and we can't change the signature without also replacing django internals. See e.g. https://github.com/django/django/blob/8093aaa8ff9dd7386a069c6eb49fcc1c5980c033/django/db/models/sql/compiler.py#L1084.

django_spanner/operations.py
```diff
@@ -133,13 +251,32 @@ def get_db_converters(self, expression):
...
     def convert_binaryfield_value(self, value, expression, connection):
+        """Convert a binary field to Cloud Spanner.
```
c24t, line 257: AFAICT these docstrings are backwards: `value` comes from python-spanner, and we're converting from one `bytes` (or ["bytes-like object"](https://docs.python.org/3/glossary.html#term-bytes-like-object)) it into another: from a base64-encoded `bytes` returned by spanner to a non-base64-encoded `bytes` for django, since that's the internal type for django `BinaryField`s.

django_spanner/operations.py
```diff
@@ -133,13 +251,32 @@ def get_db_converters(self, expression):
...
         if value is None:
             return value
         # Cloud Spanner stores bytes base64 encoded.
         return b64decode(value)
 
     def convert_datetimefield_value(self, value, expression, connection):
+        """Convert a date and time field to Cloud Spanner.
+
+        :type value: #TODO
+        :param value: A binary field value.
```
c24t, line 275: Copy/paste?

c24t (approved):

PR #567, opened 2020-12-15
chore: Add missing py3.8 classifier

c24t:
We're already testing against 3.8, this PR just adds the version classifier to setup.py.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -55,6 +55,7 @@ setup(
         "Programming Language :: Python :: 3",
         "Programming Language :: Python :: 3.6",
         "Programming Language :: Python :: 3.7",
+        "Programming Language :: Python :: 3.8",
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -55,6 +55,7 @@
         "Programming Language :: Python :: 3",
         "Programming Language :: Python :: 3.6",
         "Programming Language :: Python :: 3.7",
+        "Programming Language :: Python :: 3.8",
```
crwilcox, line 58: is there a reason not to go ahead and add 3.9 as well? (also to tests)

crwilcox (approved):

setup.py
c24t: Good call: https://github.com/googleapis/python-spanner-django/pull/568.

PR #569, opened 2020-12-16
test: use parallel workflows to run Django tests

IlyaFaer:


IlyaFaer:
@c24t, take a look at this. I think we finally can see all the Django tests executed. I had a suspicion that we never actually runned all the tests. I've taken a look through every worker in [this flow](https://source.cloud.google.com/results/invocations/25818b60-194a-475d-8c72-cc403da15908/targets/cloud-devrel%2Fclient-libraries%2Fpython%2Fgoogleapis%2Fpython-spanner-django%2Fpresubmit%2Fpresubmit/log) (September 14), when we didn't yet touch anything in the testing system. So, two workers were not actually doing anything, and the total number of executed tests across all the workers was ~1300.

Now, when I'm running eight workflows, each with its own emulator and list of apps to test, we have several times more tests executed, much longer checks and more errors - those were actually ignored previously. That requires to do some conclusions.

.kokoro/build.sh
```diff
@@ -54,6 +55,8 @@ export RUNNING_SPANNER_BACKEND_TESTS=1
 export USE_SPANNER_EMULATOR=0
 
 pip3 install .
+pip3 install -e 'git+https://github.com/q-logic/python-spanner.git@django_test_fixes#egg=google-cloud-spanner'
+pip3 install -e 'git+https://github.com/q-logic/python-spanner-django.git@dj_tests_against_emulator#egg=django-google-spanner'
```
c24t, line 59: Do you still need to install from this branch? I see it still has `add_dummy_where`?

.github/workflows/django_tests_against_emulator0.yml
```diff
@@ -0,0 +1,32 @@
...
+        ports:
+          - 9010:9010
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+      - name: Setup Python
+        uses: actions/setup-python@v2
+        with:
+          python-version: 3.8
```
c24t, line 23: We're using python 3.8 in emulator tests but 3.6 in `django_test_suite.sh`?

django_spanner/base.py
```diff
@@ -111,7 +113,9 @@ def instance(self):
         :rtype: :class:`~google.cloud.spanner_v1.instance.Instance`
         :returns: A new instance owned by the existing Spanner Client.
         """
-        return spanner.Client().instance(self.settings_dict["INSTANCE"])
+        return spanner.Client(
+            project=os.environ["GOOGLE_CLOUD_PROJECT"]
```
c24t, line 117: IIRC the client should automatically use the active gcloud project (`gcloud config get-value project`). I think this change is fine for now, and see that other googleapis libraries do something similar, but it could cause problems if the env var and active gcloud environment don't match.

create_test_instance.py
```diff
@@ -0,0 +1,33 @@
...
+    "SPANNER_EMULATOR_HOST"
+):  # otherwise instance will be created by parallelize_tests
+    project = os.getenv(
+        "GOOGLE_CLOUD_PROJECT",
+        os.getenv("PROJECT_ID", "emulator-test-project"),
+    )
+
+    client = Client(project=project)
+
+    config = f"{client.project_name}/instanceConfigs/regional-us-central1"
```
c24t, line 29: I'm surprised you need the instance config, what does https://github.com/googleapis/python-spanner-django/pull/569/commits/720ef6a77d75a2a2a622e298eea42eb78d2efe6d do?

django_test_suite.sh
```diff
@@ -4,20 +4,45 @@
...
+
+if [ $SPANNER_EMULATOR_HOST != 0 ]
+then
+    pip3 install .
+    pip3 install -e 'git+https://github.com/q-logic/python-spanner.git@django_test_fixes#egg=google-cloud-spanner'
+    pip3 install -e 'git+https://github.com/q-logic/python-spanner-django.git@dj_tests_against_emulator#egg=django-google-spanner'
+fi
+
+export DJANGO_TESTS_DIR="django_tests_dir"
+mkdir -p $DJANGO_TESTS_DIR && git clone --depth 1 --single-branch --branch spanner-2.2.x https://github.com/timgraham/django.git $DJANGO_TESTS_DIR/django
```
c24t, line 23: Another reminder, we have to fix this in https://github.com/googleapis/python-spanner-django/issues/485.

c24t (commented):
Great work splitting the emulator and non-emulator tests and getting to a green build.

Some comments and possible cleanup work, but once the `pip install -e git...` lines are removed I think we should be ready to merge.

django_test_suite.sh
```diff
@@ -4,20 +4,40 @@
...
+
+sudo apt-get update -y
+sudo apt-get install -y libmemcached-dev
+
+# Disable buffering, so that the logs stream through.
+export PYTHONUNBUFFERED=1
+
+pip3 install .
+
+export DJANGO_TESTS_DIR="django_tests_dir"
```
c24t, line 17: Looks like a lot of this file is copied from `.kokoro/build.sh`. Doesn't this mean that e.g. django is installed twice?

create_test_instance.py
IlyaFaer: It doesn't want to create an instance without minimal configurations, so we're usually creaing instances this way

django_test_suite.sh
IlyaFaer: Yeah, that's true. Moved installing Django under the if-statement

.kokoro/build.sh
```diff
@@ -54,6 +55,8 @@ export RUNNING_SPANNER_BACKEND_TESTS=1
 export USE_SPANNER_EMULATOR=0
 
 pip3 install .
+pip3 install -e 'git+https://github.com/q-logic/python-spanner.git@django_test_fixes#egg=google-cloud-spanner'
```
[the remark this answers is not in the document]
IlyaFaer: I've tried many variants, but it looks like until all the workflow changes will appear in master branch, this line will be required. In all the other ways of installing this package it's not installed properly, or doesn't contain the needed changes.

.kokoro/build.sh
IlyaFaer: We can safely drop only the first install line, and I've already did it.

.github/workflows/django_tests_against_emulator0.yml
IlyaFaer: I think it's done this way because version of kokoro Linux includes Python 3.6. But in workflows we can set Python fast and clean.

django_test_suite.sh
```diff
@@ -4,20 +4,45 @@
...
+# Disable buffering, so that the logs stream through.
+export PYTHONUNBUFFERED=1
+
+export DJANGO_TESTS_DIR="django_tests_dir"
+mkdir -p $DJANGO_TESTS_DIR
+
+if [ $SPANNER_EMULATOR_HOST != 0 ]
+then
+    pip3 install .
+    pip3 install -e 'git+https://github.com/q-logic/python-spanner-django.git@dj_tests_against_emulator#egg=django-google-spanner'
```
c24t, line 21: Reminder for later: remove this and DRY up `.kokoro/build.sh`.

c24t (approved):
Approving at the last green build rev, thanks for the long and hard work getting these tests to pass @IlyaFaer.

PR #573, opened 2021-01-07
fix: fix from-scratch tutorial

HemangChothani:


c24t (approved):
Looks good, but as with #574 these instructions could benefit from a rewrite.

PR #574, opened 2021-01-11
fix: fix healthchecks app tutorial

HemangChothani:


c24t (approved):
LGTM. There's still some editing to do here, but this is a clear improvement.

PR #578, opened 2021-01-22

HemangChothani:
First approach to implement override django autocommit to spanner db-api


HemangChothani:
Closing this PR as found better solution and opened a PR https://github.com/googleapis/python-spanner-django/pull/583

PR #579, opened 2021-01-22

HemangChothani:
Second approach to override autocommit to spanner ab-api

- For that need to change constructor of [connection](https://github.com/googleapis/python-spanner/blob/4ef793c9cd5d6dec6e92faf159665e11d63762ad/google/cloud/spanner_dbapi/connection.py#L50)  class and add `autcommit` key word argument and set connection.autocommit according to that.


HemangChothani:
Closing this PR as found better solution and opened a PR https://github.com/googleapis/python-spanner-django/pull/583

PR #583, opened 2021-01-28

HemangChothani:
It is the same patch that used in `SQLite3`, it works fine with the python-spanner-django as well.


HemangChothani:
Kokoro failed due to `rpc error: code = ResourceExhausted desc = Project 1065521786570 cannot add 1 nodes in region us-west1` error which is not related to this PR.

c24t (approved):
LGTM, I confirmed this works (with autocommit on) with the [django tutorial](https://docs.djangoproject.com/en/2.2/intro/tutorial01/) using cloud spanner and the emulator.

Thanks @HemangChothani!

PR #591, opened 2021-02-01
refactor: Erase excess code in build scripts

IlyaFaer:


IlyaFaer:
@c24t, I'm figuring now how we can safely drop parallelizing script - caught my eyes on several excess things that can be cleaned up right now.

c24t (approved):

PR #592, opened 2021-02-01
test: Remove golang test runner

IlyaFaer:


run_testing_worker.py
```diff
@@ -0,0 +1,76 @@
...
+
+instance_name, instance = create_test_instance()
+
+print(
+    """DJANGO_TEST_APPS="{apps}" SPANNER_TEST_INSTANCE={instance} bash ./django_test_suite.sh""".format(
+        apps=" ".join(test_apps), instance=instance_name
+    )
+)
+
+os.system(
```
c24t, line 70: Using `subprocess.Popen` here instead would let you print the command with [`process.args`](https://docs.python.org/3/library/subprocess.html#subprocess.CompletedProcess.args) (instead of printing above), and IIRC has a few other benefits over `os.system`.

run_testing_worker.py
```diff
@@ -0,0 +1,76 @@
...
+    created_op = instance.create()
+    created_op.result(120)  # block until completion
+    return name, instance
+
+
+worker_index = int(os.getenv("DJANGO_WORKER_INDEX", 0))
+worker_count = int(os.getenv("DJANGO_WORKER_COUNT", 1))
+
+if worker_index > worker_count:
+    print(
```
c24t, line 37: We might want to replace these `print` lines with `logger.info`, `logger.error` etc. so the script can set the verbosity instead of always printing each line.

run_testing_worker.py
```diff
@@ -0,0 +1,76 @@
...
+
+def create_test_instance():
+    project = os.getenv(
+        "GOOGLE_CLOUD_PROJECT",
+        os.getenv("PROJECT_ID", "emulator-test-project"),
+    )
+
+    client = Client(project=project)
+
+    config = f"{client.project_name}/instanceConfigs/regional-us-central1"
```
c24t, line 23: Nit: make the region a constant.

run_testing_worker.py
```diff
@@ -0,0 +1,76 @@
...
+
+    name = "spanner-django-test-{}".format(str(int(time.time())))
+
+    instance = client.instance(name, config)
+    created_op = instance.create()
+    created_op.result(120)  # block until completion
```
c24t, line 29: Do we still need this delay?

run_testing_worker.py
```diff
@@ -0,0 +1,76 @@
+# Copyright 2020 Google LLC.
```
c24t, line 1:
```suggestion
#!/usr/bin/env python3

# Copyright 2020 Google LLC.
```

And make the file executable.

run_testing_worker.py
```diff
@@ -0,0 +1,76 @@
...
+    instance = client.instance(name, config)
+    created_op = instance.create()
+    created_op.result(120)  # block until completion
+    return name, instance
+
+
+worker_index = int(os.getenv("DJANGO_WORKER_INDEX", 0))
+worker_count = int(os.getenv("DJANGO_WORKER_COUNT", 1))
+
+if worker_index > worker_count:
```
c24t, line 36: Since index is 0-indexed shouldn't this be `>=`? This would also catch `worker_count == 0`.

run_testing_worker.py
```diff
@@ -0,0 +1,76 @@
...
+    print(
+        "worker_index (wi) > worker_count (wc)".format(
+            wi=worker_index, wc=worker_count,
+        )
+    )
+    exit()
```
c24t, line 42:
```suggestion
    exit(1)
```

Same suggestion below, non-zero code for exit on error.

c24t (commented):
I left some comments, very glad to replace the binary test runner. Let's see if all the tests get run and pass.

run_testing_worker.py
IlyaFaer: Yeah, previously it was about 30 secs, but sometimes the test are failing, because instances was not created within the given time.

run_testing_worker.py
```diff
@@ -0,0 +1,76 @@
...
+
+test_apps = all_apps[start_index:end_index]
+print("test apps: ", test_apps)
+
+if not test_apps:
+    exit()
+
+delay = random.randint(10, 60)
+print("creating instance with delay: {} seconds".format(delay))
+time.sleep(delay)
```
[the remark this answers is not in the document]
IlyaFaer: Yeah, as we have several workers we're getting an error because of fast instance creation from time to time.

c24t (approved):

PR #611, opened 2021-04-22
feat: added unit test with coverage of 68 %

vi3k6i5:
feat: added unit test with coverage of 60 %


tests/unit/django_spanner/models.py
```diff
@@ -0,0 +1,64 @@
+"""
```
larkee, line 1: Test files also need the copyright header

tests/unit/django_spanner/models.py
```diff
@@ -0,0 +1,64 @@
...
+Different models used for testing django-spanner code.
+"""
+import os
+from django.db import models
+import django
+from django.db.models import Transform
+from django.db.models import CharField, TextField
```
larkee, line 8: It seems odd to have these imports when you already have: `from django.db import models`
I would remove these and just use `models.Transform`, `models.CharField`, etc.

tests/unit/django_spanner/models.py
vi3k6i5: added copyright header.

tests/unit/django_spanner/models.py
vi3k6i5: cleaned it up.

tests/unit/django_spanner/models.py
```diff
@@ -0,0 +1,67 @@
+# Copyright 2020 Google LLC
```
c24t, line 1: FYI new files should be 2021 now

tests/unit/django_spanner/test_compiler.py
```diff
@@ -0,0 +1,193 @@
...
+class TestUtils(SimpleTestCase):
+    settings_dict = {"dummy_param": "dummy"}
+
+    def _get_target_class(self):
+        from django_spanner.base import DatabaseWrapper
+
+        return DatabaseWrapper
+
+    def _make_one(self, *args, **kwargs):
+        return self._get_target_class()(*args, **kwargs)
```
c24t, line 24: Since every test method uses a connection [`TestCase.setUp`](https://docs.python.org/3/library/unittest.html#unittest.TestCase.setUp) might be a better fit here.

I'm not sure what benefit we get from `_make_one` and `_get_target_class` here.

tests/unit/django_spanner/test_compiler.py
```diff
@@ -0,0 +1,193 @@
...
+
+from django.test import SimpleTestCase
+from django.core.exceptions import EmptyResultSet
+from django.db.utils import DatabaseError
+from django_spanner.compiler import SQLCompiler
+from django.db.models.query import QuerySet
+from .models import Number
+
+
+class TestUtils(SimpleTestCase):
```
c24t, line 15: How come these classes are named `TestUtils` instead of `TestCompiler`, `TestExpressions`, etc?

tests/unit/django_spanner/test_expressions.py
```diff
@@ -0,0 +1,62 @@
...
+class TestUtils(SimpleTestCase):
+    settings_dict = {"dummy_param": "dummy"}
+
+    def _get_target_class(self):
+        from django_spanner.base import DatabaseWrapper
+
+        return DatabaseWrapper
+
+    def _make_one(self, *args, **kwargs):
+        return self._get_target_class()(*args, **kwargs)
```
c24t, line 22: Since this is repeated in every test class I'd recommend moving it into a spanner-specific base test class.

tests/unit/django_spanner/models.py
```diff
@@ -0,0 +1,67 @@
...
+# license that can be found in the LICENSE file or at
+# https://developers.google.com/open-source/licenses/bsd
+"""
+Different models used for testing django-spanner code.
+"""
+import os
+from django.db import models
+import django
+
+# Load django settings before loading dhango models.
```
c24t, line 13:
```suggestion
# Load django settings before loading django models.
```

tests/unit/django_spanner/models.py
```diff
@@ -0,0 +1,67 @@
...
+os.environ["DJANGO_SETTINGS_MODULE"] = "tests.settings"
+django.setup()
```
c24t, line 15: Would using [`SimpleTestCase.settings`](https://docs.djangoproject.com/en/3.2/topics/testing/tools/#django.test.SimpleTestCase.settings) let you avoid clobbering the settings here?

c24t (commented):
A few comments on test organization and style, but the tests themselves look great and this is a big obvious improvement.

I'll have to take another pass to check the test logic in more detail, but the tests I've checked so far all LGTM.

@larkee may have some insight on test strategy from python-spanner.

tests/unit/django_spanner/models.py
vi3k6i5: updated to 2021 in all files added in 2021.

tests/unit/django_spanner/test_compiler.py
vi3k6i5: Copy paste, my bad, cleaned it up.

tests/unit/django_spanner/models.py
vi3k6i5: fixed.

tests/unit/django_spanner/test_expressions.py
vi3k6i5: Done.

tests/unit/django_spanner/models.py
vi3k6i5: SimpleTestCase came with some overhead of setup and cleaning, got rid of it and using `unittest.TestCase` now.

tests/unit/django_spanner/test_expressions.py
vi3k6i5: Done.

c24t (approved):
Tests look great, thanks for the updates

PR #613, opened 2021-04-25
feat: move migrations test modules to run against different emulator

vi3k6i5:
feat: move migrations test modules to run against different emulator to distribute time taken by system test 2


larkee (approved):
minor nit: title can be shortened to `feat: move migrations test modules to run against different emulator` if you add the explanation to the PR description :+1:

PR #621, opened 2021-05-09
ci: removes nox run on all kokoro workers and moves it to github actions

vi3k6i5:
ci: removed nox run on all kokoro workers and moved it to github actions.

Currently all kokoro workers run nox, leading to total runtime increase. So moved them out of the kokoro worker tasks and added it to github actions.


larkee (commented):
This PR does not affect the performance of the library so it should not be tagged with `perf`. It is changing the CI setup so it should use `ci`.

The rest of it: LGTM

larkee (approved):

PR #623, opened 2021-05-13
ci: add system tests

### LINKED ISSUE #624: ci: Add system tests similar to python spanner.
ci: Add system tests similar to python spanner client.

vi3k6i5:
ci: Adding system tests.

Currently tests are split into 3 parts:
1. Unit tests: Run by using command `nox -s unit`. This test is used for library code testing and does not interact with spanner or emulator.
2. Github workflows: These run the django test modules one by one using `sh django_test_suite.sh`. These tests run agains't the emulator. 
3. Kokoro: Some of the django test modules are very time consuming and thus to reduce the total runtime of tests we use kokoro to run a few select tests cases agains't actual spanner. Command used for this is `run_testing_worker.py` and modules tested are listed in [https://github.com/googleapis/python-spanner-django/blob/master/django_test_apps.txt](django_test_apps.txt).

We are introducing a new test called system tests.
These tests are similar to python spanner library system tests in the sense that they would run on [github workflow](https://github.com/googleapis/python-spanner-django/blob/master/.github/workflows/integration-tests-against-emulator.yml). These tests would actually be interacting with emulator and using the library they would perform operations like creating table, inserting data, running queries etc. 

fixes #624


larkee (approved):
LGTM as long as the tests pass :+1:

PR #633, opened 2021-05-20
feat: Add support for open telemetry

vi3k6i5:
Add support for open telemetry tracing if open telemetry is installed.
Added the same as extra dependency.

Also added unit tests for the same.

Follow up PR to add tracing in schema with related checks in unit tests. https://github.com/googleapis/python-spanner-django/pull/648

Fixes #633 🦕


tests/_helpers.py
```diff
@@ -0,0 +1,79 @@
...
+    provider.add_span_processor(SimpleSpanProcessor(get_test_ot_exporter()))
+    _TEST_OT_PROVIDER_INITIALIZED = True
+
+
+class OpenTelemetryBase(unittest.TestCase):
+    @classmethod
+    def setUpClass(cls):
+        if HAS_OPENTELEMETRY_INSTALLED:
+            use_test_ot_exporter()
+            cls.ot_exporter = get_test_ot_exporter()
```
c24t, line 57: I know I mentioned it in https://stackoverflow.com/q/67826617 too, but one downside of solving the problem this way is that all instances of the test will share the same exporter and so these tests will have concurrency problems.

I think this is still the right solution here until OT makes this easier, or unless you decide this is a big enough problem to dig into the internals of OT context.

tests/unit/django_spanner/simple_test.py
```diff
@@ -7,13 +7,16 @@
...
 from django_spanner.operations import DatabaseOperations
-from unittest import TestCase
+
+# from unittest import TestCase
+from tests._helpers import OpenTelemetryBase
 import os
 
 
-class SpannerSimpleTestClass(TestCase):
+class SpannerSimpleTestClass(OpenTelemetryBase):
```
c24t, line 16: It seems like most test classes won't also test tracing, any reason not to make a separate test class for tracing? E.g. `SpannerTracingTestClass`?

And what about making the django tests use [django's `TestCase`](https://docs.djangoproject.com/en/2.2/topics/testing/overview/) instead of `unittest.TestCase`?

Something like:

```python
from django.test import TestCase

class SpannerSimpleTestCase(TestCase)
    ...

class SpannerTracingCase(SpannerSimpleTestCase, OpenTelemetryBase)
    ...
```

django_spanner/_opentelemetry_tracing.py
```diff
@@ -0,0 +1,52 @@
...
+
+    HAS_OPENTELEMETRY_INSTALLED = True
+except ImportError:
+    HAS_OPENTELEMETRY_INSTALLED = False
+
+
+@contextmanager
+def trace_call(name, connection, extra_attributes=None):
+    if not HAS_OPENTELEMETRY_INSTALLED or not connection:
+        # Empty context manager. Users will have to check if the generated value is None or a span
```
c24t, line 25: Nit: here and below, comment wrapping

c24t (approved):
The way we're testing OT is kind of surprising, but that's largely OT's fault for being difficult to test. A few minor comments, but LGTM.

larkee (commented):
So this adds the `trace_call` method but does not add any calls. This means there is no tracing being added to the library. Is the intention for users to add trace calls themselves?

For the Spanner Client and the SQLAlchemy connector, we added tracing within the libraries. See [here](https://github.com/googleapis/python-spanner/pull/107) and [here](https://github.com/cloudspannerecosystem/python-spanner-sqlalchemy/pull/51) for more context.

tests/_helpers.py
vi3k6i5: For now the tests are working so not digging deeper at the moment, if I face any issue with parallelisation of tests will try to resolve it.

tests/unit/django_spanner/simple_test.py
vi3k6i5: the [django TestCase](https://docs.djangoproject.com/en/2.2/topics/testing/overview/) inherits from `TransactionTestCase` which inherits from `SimpleTestCase` which does database connection check, and we don't connect to db for our unit tests so it's simpler to not use it at least for unit tests.

django_spanner/_opentelemetry_tracing.py
vi3k6i5: fixed.

vi3k6i5:
> So this adds the `trace_call` method but does not add any calls. This means there is no tracing being added to the library. Is the intention for users to add trace calls themselves?
> 
> For the Spanner Client and the SQLAlchemy connector, we added tracing within the libraries. See [here](https://github.com/googleapis/python-spanner/pull/107) and [here](https://github.com/cloudspannerecosystem/python-spanner-sqlalchemy/pull/51) for more context.

So that will have to be a follow up PR, since this PR was already crossing 300 lines, I didn't add it here.

PR #641, opened 2021-05-26
fix: remove delete of instance if it already exists, as it might be in use by another test

vi3k6i5:
fix: remove delete of instance if it already exists, as it might be in use by another test
Also include WORKER_INDEX in instance name to reduce chance of clash across workers.

### DIFF
--- a/run_testing_worker.py
+++ b/run_testing_worker.py
@@ -34,10 +34,6 @@ class TestInstance:
         )
 
         self._instance = client.instance(name, config)
-        if self._instance.exists():
-            # If test instance already exists first delete it and then create.
-            self._instance.delete()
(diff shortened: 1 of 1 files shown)


run_testing_worker.py
```diff
@@ -14,6 +14,8 @@
 from google.cloud.spanner_v1 import Client
 
 REGION = "regional-us-central1"
+WORKER_INDEX = int(os.getenv("DJANGO_WORKER_INDEX", 0))
```
thiagotnunes, line 17: I think you could use a `UUID` here to (practically) guarantee no clashes will happen.

thiagotnunes (approved):

PR #648, opened 2021-06-16
feat: Add open telemetry trace in schema and related unit tests

vi3k6i5:
Add tracing in schema with related checks in unit tests. 

Fixes #633 🦕


larkee (commented):
LGTM, although I'm not very familiar with Django usage.

For `CloudSpannerDjango.alter_field`, I noticed that the method is `_alter_field`. Should the trace reflect this? Do users call this directly?

vi3k6i5:
> _alter_field

users are not going to call this method directly, this trace is added more for keeping track. Wherever calls to backend takes place then how long those calls are taking.

asthamohta (approved):

PR #658, opened 2021-06-21
fix: Bump version number after 2.2.1b1 release

vi3k6i5:
Like #596, bump the version number following the 2.2.1b1 release to avoid using the same version number in development.

This has to be done manually because Release-please doesn't understand our non-standard versioning.


vi3k6i5:
Changes already merged in #657

PR #721, opened 2021-10-11
feat: enable support for `get_key_columns` and cleanup tests with unknown failures to specific failures.

vi3k6i5:
feat: enable support for `get_key_columns` and cleanup tests with unknown failures to specific failures.

Note: ON DELETE CASCADE is not supported still for Foreign key, so this feature is not full fledged yet.


asthamohta (approved):
LGTM

Issue #742, opened 2021-12-15
Autofield patching doesn't work for multi-database migration against MySQL


skyl:
I am trying to keep all of my existing tables in MySQL but add "spanner" as a second database for certain apps.

`DATABASES` looks something like this:

```
{'default': {'ENGINE': 'django.db.backends.mysql',
  'NAME': 'dbname',
  'USER': 'username',
  'PASSWORD': 'FOOBAR',
  'HOST': '127.0.0.1',
  'PORT': '3306',
  'CONN_MAX_AGE': 240,
  'ATOMIC_REQUESTS': False,
  'AUTOCOMMIT': True,
  'OPTIONS': {},
  'TIME_ZONE': None,
  'TEST': {'CHARSET': None, 'COLLATION': None, 'NAME': None, 'MIRROR': None}},

 'spanner': {'ENGINE': 'django_spanner',
  'PROJECT': 'my-gcp-project',
  'INSTANCE': 'manual-test0',
  'NAME': 'test1',
  'ATOMIC_REQUESTS': False,
  'AUTOCOMMIT': True,
  'CONN_MAX_AGE': 0,
  'OPTIONS': {},
  'TIME_ZONE': None,
  'USER': '',
  'PASSWORD': '',
  'HOST': '',
  'PORT': '',
  'TEST': {'CHARSET': None, 'COLLATION': None, 'NAME': None, 'MIRROR': None}}}
```

I have a router that is much like the one in the multi-db docs that puts just 1 table(app) in spanner. 

```
class SpannerAppRouter:
    """
    A router to control all database operations on models in the
    myapp application to spanner
    """
    spanner_app_labels = {'myapp'}

    def db_for_read(self, model, **hints):

        if model._meta.app_label in self.spanner_app_labels:
            return 'spanner'
        return None

    def db_for_write(self, model, **hints):
        if model._meta.app_label in self.spanner_app_labels:
            return 'spanner'
        return None

    def allow_relation(self, obj1, obj2, **hints):
        """
        Allow relations if all models involved are in spanner
        """
        if all([
            obj1._meta.app_label in self.spanner_app_labels,
            obj2._meta.app_label in self.spanner_app_labels
        ]):
            return True
        return None

    def allow_migrate(self, db, app_label, model_name=None, **hints):
        """
        Make sure spanner tables only appear in the 'spanner' database
        and spanner tables don't appear in the default
        """
        # print(db, app_label, model_name)
        if app_label in self.spanner_app_labels:
[...]

PR #771, opened 2022-06-17
fix: don't override primary key for non-Spanner databases used by the…

IlyaFaer:
… same Django instance

Towards #742


IlyaFaer:
A lot of tests failed, because Django doesn't consider the `max_query_params` attribute. Feature-requesting this with the Django team.

IlyaFaer:
During fixing the `AutoField` patch, a bunch of errors were fixed in order to make the tests pass:

- Test are failing because during a test database installation Django is trying to execute a query with more than 950 parameters (is not possible on Spanner)
- `TZ_INFO` variable default value is going to be changed in the next versions of Django. A warning was added to notify users - the tests were failing because of it.

PR #772, opened 2022-07-04
chore: use stable django versions for tests hosted on python-spanner-django

ansh0l:
chore: use django stable versions for tests hosted on python-spanner-django

### DIFF
--- a/.github/sync-repo-settings.yaml
+++ b/.github/sync-repo-settings.yaml
@@ -9,10 +9,6 @@ branchProtectionRules:
   requiredStatusCheckContexts:
     - 'cla/google'
     - 'OwlBot Post Processor'
-    - 'system-tests (3.6)'
-    - 'system-tests (3.7)'
-    - 'system-tests (3.8)'
-    - 'system-tests (3.9)'
     - 'system-tests'
     - 'Kokoro'
 permissionRules:
--- a/django_test_suite.sh
+++ b/django_test_suite.sh
@@ -18,7 +18,7 @@ mkdir -p $DJANGO_TESTS_DIR
 if [ $SPANNER_EMULATOR_HOST != 0 ]
 then
     pip3 install .
-    git clone --depth 1 --single-branch --branch "spanner/stable/2.2.x" https://github.com/c24t/django.git $DJANGO_TESTS_DIR/django
+    git clone --depth 1 --single-branch --branch "django/stable/2.2.x" https://github.com/googleapis/python-spanner-django.git $DJANGO_TESTS_DIR/django
 fi
 
 # Install dependencies for Django tests.
--- a/django_test_suite.sh
+++ b/django_test_suite.sh
@@ -18,7 +18,7 @@ mkdir -p $DJANGO_TESTS_DIR
 if [ $SPANNER_EMULATOR_HOST != 0 ]
 then
     pip3 install .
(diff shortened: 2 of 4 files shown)


asthamohta (approved):

IlyaFaer (approved):
LGTM
There are several errors, caused by Django (not us), see this comment:
https://github.com/googleapis/python-spanner-django/pull/771#issuecomment-1170880766
Thus, I guess, this one can be merged despite the failures. I'll push max params number hit fix to the new forks.

IlyaFaer:
Also I've created an issue for another failure we're experiencing in Django:
https://github.com/googleapis/python-api-core/issues/407

ansh0l:
@IlyaFaer Some of the failing presubmits are required ones. I don't think skipping them is a good idea, Lets connect to understand what specifically is going wrong with them.

IlyaFaer:
@ansh0l, to fix errors we need to start using new Django forks. We can't fix the failures in `spanner_django` and it's not convenient to push fixes into personal Django forks of people, who are no longer working in Spanner. That's why it was decided to create forks in `googleapis` account. It's the first step of fixing the failures.

ansh0l:
@IlyaFaer : What if we pushed in the updated Django stable releases as a separate branch on this project (python-spanner-django) itself, and made them protected branch? This is what this PR is essentially trying to do.

IlyaFaer:
@ansh0l, not sure what it'll give us. These required tests are failing in all the new PRs already, and this PR is just switching forks. It's not directly related, but we'll have to sync about four fixes in three repositories to see the tests passing (and probably it'll still be not enough, because one of the errors is failing in the `api_core`). I'd better merge them one by one, just checking that the error is fixed (even if another one, not related to the fix, is failing), delivering users at least some fixes, instead of holding them all.

Still, it's your decision, I'm a guest-dev here

IlyaFaer:
Now you'll have to change branches in your PR to use fixes for Django 2.2 and 3.2: #773 and #774
Also you need to push into this PR this fix:

Untitled (1)

Than a multi-regions db fix can be applied from this PR:
https://github.com/googleapis/python-spanner-django/pull/771/files

And then there is still an import error, for which, as I said, I've already created an issue in `api_core`:
https://github.com/googleapis/python-api-core/issues/407

If all of it will be gathered correctly in your PR, well, maybe we'll see the tests passing fine.

IlyaFaer:
@asthamohta, @ansh0l, oh, no! The fork pushed into `django/stable/2.2.x` is ~3000 commits behind the one I'm using. And I'm thinking why this TZ_INFO error don't show up in this PR? It's very old. And fix for params num is not working also, because the method significantly changed:

image

Looks like tests were redirected and the newer PRs into the fork were made into the `stable/2.2.x`, which is far behind `main`, which was supposed to be used.

We really should think about how to get rid of using forks ASAP - they are creating confusion during the whole history of the package.

ansh0l:
Thanks Ilya, sounds fair. Essentially, I've pushed the same commits as branches, that were being referenced in the previous forks, so that everything else in the environment remains the same. 

I agree that the forks are not an ideal solution.

Should we start relying on the django repository directly for these tests? Also, can you point me to the specific commit SHA for Django that you have against which the broken tests are working fine?

IlyaFaer:
@asthamohta. @ansh0l, okay, the main tests passed. Fixes are in this two PRs: https://github.com/googleapis/python-spanner-django/pull/777, https://github.com/googleapis/python-spanner-django/pull/776 (I've switched this PR to use those branch and will revert it as soon as the PRs are merged)

PR #775, opened 2022-07-07
fix: decrease query param chunk size

IlyaFaer:


django/db/models/query.py
```diff
@@ -1204,7 +1216,7 @@ def _batched_insert(self, objs, fields, batch_size, ignore_conflicts=False):
         batch_size = (batch_size or max(ops.bulk_batch_size(fields, objs), 1))
         inserted_ids = []
         bulk_return = connections[self.db].features.can_return_ids_from_bulk_insert
-        for item in [objs[i:i + batch_size] for i in range(0, len(objs), batch_size)]:
+        for item in [objs[i : i + 100] for i in range(0, len(objs), 100)]:
```
IlyaFaer, line 1219: `objs` here are not just rows, but Django high-level objects, which can be unwrapped into several rows.

asthamohta (approved):

PR #780, opened 2022-07-19
fix: override AutoField default value only for Spanner

### LINKED ISSUE #742: Autofield patching doesn't work for multi-database migration against MySQL

IlyaFaer:
Fixes #742

### DIFF
--- a/django_spanner/__init__.py
+++ b/django_spanner/__init__.py
@@ -15,6 +15,7 @@ from uuid import uuid4
 import pkg_resources
 from google.cloud.spanner_v1 import JsonObject
 from django.db.models.fields import (
+    NOT_PROVIDED,
     AutoField,
     Field,
 )
@@ -61,7 +62,12 @@ def gen_rand_int64():
 def autofield_init(self, *args, **kwargs):
     kwargs["blank"] = True
     Field.__init__(self, *args, **kwargs)
-    self.default = gen_rand_int64
+
+    if (
+        django.db.connection.settings_dict["ENGINE"] == "django_spanner"
+        and self.default == NOT_PROVIDED
+    ):
+        self.default = gen_rand_int64
 
 
 AutoField.__init__ = autofield_init
--- a/tests/unit/django_spanner/test_schema.py
+++ b/tests/unit/django_spanner/test_schema.py
@@ -6,9 +6,10 @@
 
 
 from .models import Author
-from django.db import NotSupportedError
+from django.db import NotSupportedError, connection
 from django.db.models import Index
-from django.db.models.fields import IntegerField
+from django.db.models.fields import AutoField, IntegerField
+from django_spanner import gen_rand_int64
 from django_spanner.schema import DatabaseSchemaEditor
 from tests._helpers import HAS_OPENTELEMETRY_INSTALLED
 from tests.unit.django_spanner.simple_test import SpannerSimpleTestClass
@@ -404,3 +405,31 @@ class TestUtils(SpannerSimpleTestClass):
             new_field.set_attributes_from_name("author_num")
             with self.assertRaises(NotSupportedError):
                 schema_editor.alter_field(Author, old_field, new_field)
+
+    def test_autofield_no_default(self):
+        """Spanner, default is not provided."""
+        field = AutoField(name="field_name")
+        assert gen_rand_int64 == field.default
+
+    def test_autofield_default(self):
+        """Spanner, default provided."""
+        mock_func = mock.Mock()
+        field = AutoField(name="field_name", default=mock_func)
+        assert gen_rand_int64 != field.default
+        assert mock_func == field.default
+
+    def test_autofield_not_spanner(self):
+        """Not Spanner, default not provided."""
+        connection.settings_dict["ENGINE"] = "another_db"
+        field = AutoField(name="field_name")
+        assert gen_rand_int64 != field.default
+        connection.settings_dict["ENGINE"] = "django_spanner"
+
+    def test_autofield_not_spanner_w_default(self):
+        """Not Spanner, default provided."""
+        connection.settings_dict["ENGINE"] = "another_db"
+        mock_func = mock.Mock()
+        field = AutoField(name="field_name", default=mock_func)
+        assert gen_rand_int64 != field.default
+        assert mock_func == field.default
+        connection.settings_dict["ENGINE"] = "django_spanner"


django_spanner/__init__.py
```diff
@@ -61,7 +62,12 @@ def gen_rand_int64():
 def autofield_init(self, *args, **kwargs):
     kwargs["blank"] = True
     Field.__init__(self, *args, **kwargs)
-    self.default = gen_rand_int64
+
+    if (
+        django.db.connection.settings_dict["ENGINE"] == "django_spanner"
+        and self.default == NOT_PROVIDED
+    ):
```
IlyaFaer, line 69: Using the INT64 random key only in case it's a Spanner database and default value/function for the auto-field is not provided. Thus, outside Spanner database it'll work as always, and for Spanner users will still be able to override the key with whatever they want.

asthamohta (approved):
Lgtm

Issue #783, opened 2022-08-09
django-spanner's autofield generation does not work for non-default spanner DBs


untitaker:
https://github.com/googleapis/python-spanner-django/pull/780 introduced a check as to whether the _default_ database's engine is spanner or not, and depending on that decides to override id generation with clientside uuid or not.

This fix is insufficient. It does improve the situation, as django-spanner will disable its monkeypatches if the default db is not spanner.

But the original issue https://github.com/googleapis/python-spanner-django/issues/742 was talking about using multiple databases, where `default` db is mysql. This is what `django.db.connection.settings_dict` refers to. The correct key to check would be `django.db.connections[???].settings_dict`.

### Steps to reproduce

1. set up spanner as a non-default database
2. see that even running basic migrations against that non-default db fails, as the inserts into the `django_migrations` table already fail.

I'll investigate how to fix. It's likely something else has to be patched entirely.

#### Environment details

  - Programming language: Python
  - OS: Doesn't matter
  - Language runtime version: 3.8
  - Package version: master

IlyaFaer:
About the next spring a new auto incrementing primary keys features is expected to be released in Spanner. It should make autofield patch unnecessary at all.

mitsuo0114:
+1. what is current status?
I've found the auto incrementing primary keys feature was introduced.
https://cloud.google.com/blog/products/databases/announcing-support-for-auto-generated-keys-in-spanner?hl=en

But still I'm facing error when migration info is inserted to `django_migrations`.
```
"id must not be NULL in table django_migrations."
```

On Cloud Spanner, following table is created automatically.
The table definition of `django_migrations` does not include configs for auto-generate-keys.
```
CREATE TABLE django_migrations (
  id INT64 NOT NULL,
  app STRING(255) NOT NULL,
  name STRING(255) NOT NULL,
  applied TIMESTAMP NOT NULL,
) PRIMARY KEY(id);;
```

I manually configured `django_migrations` with sequence like below and migration succeeded.

```
SELECT GET_NEXT_SEQUENCE_VALUE(SEQUENCE serial);

CREATE TABLE django_migrations (
  id INT64 DEFAULT (GET_NEXT_SEQUENCE_VALUE(SEQUENCE serial)),
  app STRING(255) NOT NULL,
  name STRING(255) NOT NULL,
  applied TIMESTAMP NOT NULL,
) PRIMARY KEY(id);;
```

[...]

PR #784, opened 2022-08-09
Fix support for autofields in non-default databases

untitaker:
Fix https://github.com/googleapis/python-spanner-django/issues/783

Also refactored code a bit such that version detection is no longer used
to determine whether BigAutoField exists. BigAutoField exists in Django
versions older than 3.0, and the way this is currently monkeypatched
prevents us from using any BigAutoField in Spanner with Django 2.2.


untitaker:
Turns out, neither commit in this PR is sufficient. The django migration recorder can run in any DB, so `router.db_for_write` is useless.

Need to patch the querybuilder to succeed IMO.

untitaker:
I am convinced that the `validate_autopk_value` function is the right place to hook, but that function will not be called unless there is a value for the autofield. Which would bring us back to patching models/fields.

olavloite:
Superseded by #940

PR #811, opened 2022-12-21
test: drop Python 3.6 tests

IlyaFaer:


IlyaFaer:
@asthamohta, looks like `Run Django Spanner integration tests against emulator` are organized in a bit wrong way - workflows are trying to use different versions of Python at the same time, which causes errors. Will see how to fix it.

PR #841, opened 2023-11-03
chore: update docfx minimum Python version

dandhlee:
Updates minimum Python version required for DocFX to 3.10. See https://github.com/googleapis/synthtool/pull/1891.

Python3.10 is available by default for Ubuntu22.04, which this repository uses. No installation required.

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -201,7 +201,7 @@ def docs(session):
     )
 
 
-@nox.session(python="3.9")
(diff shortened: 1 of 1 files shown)


parthea:
The failing presubmit is tracked in https://github.com/googleapis/python-spanner-django/issues/837

parthea (approved):

PR #858, opened 2024-02-26
chore: Add support for named schema

ankiaga:


django_spanner/introspection.py
```diff
@@ -38,7 +38,7 @@ class DatabaseIntrospection(BaseDatabaseIntrospection):
             FROM
                 information_schema.tables AS t
             WHERE
-                t.table_catalog = '' and t.table_schema = ''
+                t.table_catalog = '' and t.table_schema = '{schema_name}'
```
olavloite, line 41: Is the `if USE_EMULATOR` statement still needed?

django_spanner/introspection.py
```diff
@@ -76,7 +76,10 @@ def get_table_list(self, cursor):
         :rtype: list
         :returns: A list of table and view names in the current database.
         """
-        results = cursor.run_sql_in_snapshot(self.LIST_TABLE_SQL)
+        schema_name = self._get_schema_name(cursor)
+        results = cursor.run_sql_in_snapshot(
+            self.LIST_TABLE_SQL.format(schema_name=schema_name)
```
olavloite, line 81: I know it's not introduced by this PR, but if I understand it correctly, all of these methods generate a SQL statement with a literal in the string, instead of using a query parameter. That means that they are all potentially susceptible to SQL injection. Can we fix that for the queries that we are working on in this PR, and then do a follow-up PR for all other queries? (This should have a high priority)

django_spanner/introspection.py
ankiaga: Removed as the mentioned bug is closed and support has been added to emulator

django_spanner/introspection.py
ankiaga: Fixed

django_spanner/introspection.py
```diff
@@ -225,17 +225,17 @@ def get_constraints(self, cursor, table_name):
...
+            """
             SELECT
                 CONSTRAINT_NAME, COLUMN_NAME
             FROM
                 INFORMATION_SCHEMA.CONSTRAINT_COLUMN_USAGE
-               WHERE TABLE_NAME="{table}"'''.format(
-                table=quoted_table_name
-            )
+               WHERE TABLE_NAME=@table AND TABLE_SCHEMA=@schema_name""",
+            params={"table": quoted_table_name, "schema_name": schema_name},
```
olavloite, line 238: We should not use a quoted table name when it is used as a query parameter. Also; as there is no test failing due to this, I think it means that this method is also not covered by any tests. Would you mind adding a test for this method as well?

django_spanner/introspection.py
```diff
@@ -254,15 +254,14 @@ def get_constraints(self, cursor, table_name):
...
             SELECT
                 CONSTRAINT_NAME, CONSTRAINT_TYPE
             FROM
                 INFORMATION_SCHEMA.TABLE_CONSTRAINTS
             WHERE
-                TABLE_NAME="{table}"'''.format(
-                table=quoted_table_name
-            )
+                TABLE_NAME=@table AND TABLE_SCHEMA=@schema_name""",
+            params={"table": quoted_table_name, "schema_name": schema_name},
```
olavloite, line 264: Here also: The table name should not be quoted when used as a parameter. Can we have a test for this method as well?

django_spanner/introspection.py
```diff
@@ -303,14 +302,13 @@ def get_constraints(self, cursor, table_name):
...
             WHERE
-                idx.TABLE_NAME="{table}"
+                idx.TABLE_NAME=@table AND idx.TABLE_SCHEMA=@schema_name
             ORDER BY
                 idx_col.ORDINAL_POSITION
-            """.format(
-                table=quoted_table_name
-            )
+            """,
+            params={"table": quoted_table_name, "schema_name": schema_name},
```
olavloite, line 311: Same regarding quoted table name

django_spanner/introspection.py
```diff
@@ -366,10 +365,15 @@ def get_key_columns(self, cursor, table_name):
...
                 rc.CONSTRAINT_NAME = ccu.CONSTRAINT_NAME
             WHERE
-                tc.TABLE_NAME="{table}"
-            """.format(
-                table=self.connection.ops.quote_name(table_name)
-            )
+                tc.TABLE_NAME=@table AND tc.TABLE_SCHEMA=@schema_name
+            """,
+            params={
+                "table": self.connection.ops.quote_name(table_name),
```
olavloite, line 371: here also: table name should not be quoted, and the lack of a test failure to me seems like a lack of test coverage

django_spanner/introspection.py
ankiaga: Fixed as discussed

django_spanner/introspection.py
ankiaga: Fixed as discussed

django_spanner/introspection.py
ankiaga: Fixed as discussed

django_spanner/introspection.py
ankiaga: Fixed as discussed

django_spanner/introspection.py
```diff
@@ -303,14 +295,13 @@ def get_constraints(self, cursor, table_name):
             RIGHT JOIN
                 INFORMATION_SCHEMA.INDEX_COLUMNS AS idx_col
             ON
-                idx_col.INDEX_NAME = idx.INDEX_NAME AND idx_col.TABLE_NAME="{table}"
+                idx_col.INDEX_NAME = idx.INDEX_NAME AND idx_col.TABLE_NAME=@table
```
olavloite, line 298: This type of join clauses should also include the schema name. Otherwise, two indexes (or tables) that have the same name, but live in different schemas, could be joined here.

```suggestion
                idx_col.INDEX_NAME = idx.INDEX_NAME AND idx_col.TABLE_NAME=@table AND idx_col.TABLE_SCHEMA=idx.TABLE_SCHEMA
```

olavloite (approved):
LGTM, except for one small nit on a missing clause in a JOIN condition.

PR #860, opened 2024-03-11
Changes for Django 4.2. This PR is not for review as it is just used for testing the ongoing changes

ankiaga:


django_spanner/__init__.py
```diff
@@ -46,7 +46,7 @@
 USE_EMULATOR = os.getenv("SPANNER_EMULATOR_HOST") is not None
 
 # Only active LTS django versions (2.2.*, 3.2.*) are supported by this library right now.
```
jaz888, line 48: nit: (2.2.*, 3.2.*, 4.2.*)

django_spanner/__init__.py
```diff
@@ -31,7 +31,7 @@
 
 
 USING_DJANGO_3 = False
-if django.VERSION[:2] == (3, 2):
+if django.VERSION[:2] == (3, 2) or django.VERSION[:2] == (4, 2):
     USING_DJANGO_3 = True
```
jaz-la, line 35: maybe just check if django version is >=3.2 and <= 5.2, and thrown an exception otherwise?
this will drop the support of 2.2

jaz-la: also  USING_DJANGO_3 is semantically wrong once 4.2 is supported.

django_spanner/schema.py
```diff
@@ -117,14 +117,22 @@ def create_model(self, model):
...
             # allow them inline on a column.
             if field.unique and not field.primary_key:
-                self.deferred_sql.append(
-                    self._create_unique_sql(model, [field.column])
-                )
+                if USING_DJANGO_3:
+                    self.deferred_sql.append(
+                        self._create_unique_sql(model, [field])
+                    )
+                else:
```
jaz-la, line 124: this is called for both django 2.2 and 4.2?

django_spanner/schema.py
```diff
@@ -117,14 +117,22 @@ def create_model(self, model):
             # Create a unique constraint separately because Spanner doesn't
             # allow them inline on a column.
             if field.unique and not field.primary_key:
-                self.deferred_sql.append(
-                    self._create_unique_sql(model, [field.column])
-                )
+                if USING_DJANGO_3:
```
jaz-la, line 120: with django 3, originally it calls `self._create_unique_sql(model, [field.column])` and now it calls `self._create_unique_sql(model, [field])` is it actually right?

jaz-la:
given that the change is small, can we support django 5.0, and drop the support for 2.2? In this case, the USING_DJANGO_3 can also be removed?

django_spanner/__init__.py
ankiaga: Thanks for the comments but this is just a draft PR that I am using for testing for Django 4.2 so its just a hack to test it and not ready for review.

jaz-la: Thank you @ankiaga , if the effort to support 5.0 is not a lot on top of 4.2, can we also add support to 5.0?
This will allow spanner users to upgrade to django 5.0, everybody is on 3.2 currently, I prefer migrate from 3.2->5.0 in one go, instead of 3.2 -> 4.2 -> then in 2 years -> 5.2

ankiaga: We are first planning to release 4.2 and then we will take a call for supporting 5.0 as well but we are already discussing this internally

grjones: With this release, could it at least be possible to override the version check? I'm fairly certain these changes will work for 5.

jaz-la: +1 to @grjones , if we don't officially support 5.0, maybe allow users to bypass the check and take our own risk

ankiaga: Please note that these are not complete changes for Django 4.2 and I am fixing them as I am finding more issues.

Also to release 5.0 (similar to 4.2) we need to make sure that all the system tests from Django core repo works whose number is in tens of thousands. This is the main challenge to release a new version. Currently many hundreds of tests are failing as I am trying to migrate to 4.2 and root causing and fixing them takes a lot of time

grjones:
@ankiaga Thank you for all of your work on this. It means a lot. 😄

PR #865, opened 2024-04-12
feat!: Support Django 4.2

ankiaga:
## Breaking changes
- Removes support for Django 2.2
- Transaction handling in Django 4.2 is different than in Django 3.2:
  - In Django 3.2: When Django tries to manually start a transaction while the dbapi connection is in `AutoCommit=True` mode, the Spanner Django provider would __not__ start a transaction.
  - In Django 4.2: When Django wants to manually start a transaction while the dbapi connection is in `AutoCommit=True` mode, the Spanner Django provider __will start a transaction__.

## Release notes

1. Adds support for Django 4.2
2. Removes support for Django 2.2
3. Only Python versions 3.8 and higher are supported 
4. Adds support for transactions in `AutoCommit=True` mode by default for V4.2. Transactions in `AutoCommit=True` mode are by default not supported in prior versions (i.e. V3.2). 
   - To enable manual transactions in `AutoCommit=True` mode in V3.2, please set `ALLOW_TRANSACTIONS_IN_AUTO_COMMIT=True` your settings.py file. 
   - To disable transactions in `AutoCommit=True` mode in V4.2, please set `ALLOW_TRANSACTIONS_IN_AUTO_COMMIT=False` in your settings.py file.


django_spanner/__init__.py
```diff
@@ -34,22 +33,24 @@
...
+)
+from django.db.models import JSONField
 
 __version__ = pkg_resources.get_distribution("django-google-spanner").version
 
 USE_EMULATOR = os.getenv("SPANNER_EMULATOR_HOST") is not None
 
 # Only active LTS django versions (2.2.*, 3.2.*) are supported by this library right now.
-SUPPORTED_DJANGO_VERSIONS = [(2, 2), (3, 2)]
+SUPPORTED_DJANGO_VERSIONS = [(2, 2), (3, 2), (4, 2)]
```
jaz-la, line 51: Once this change it merged, will 2.2 & 3.2 still be compatible?

django_spanner/compiler.py
```diff
@@ -38,76 +39,173 @@ def get_combinator_sql(self, combinator, all):
...
-        parts = ()
-        for compiler in compilers:
-            try:
-                # If the columns list is limited, then all combined queries
-                # must have the same columns list. Set the selects defined on
-                # the query on all combined queries, if not already set.
-                if (
-                    not compiler.query.values_select
-                    and self.query.values_select
+        if USING_DJANGO_3:
```
jaz-la, line 42: This big if/else looks pretty bad:
1. they have a lot of common code that can be put in 1 place
2. the actual different logic between django 3/4 seems to be small but critical, does it make sense to provide a short comment, and minimize the length of code in this if/else

django_spanner/__init__.py
```diff
@@ -34,22 +33,24 @@
...
     USING_DJANGO_3 = True
 
-if USING_DJANGO_3:
-    from django.db.models.fields import (
-        SmallAutoField,
-        BigAutoField,
-    )
-    from django.db.models import JSONField
+USING_DJANGO_4 = False
+if django.VERSION[:2] == (4, 2):
```
jaz-la, line 37: as commented in this draft:
https://github.com/googleapis/python-spanner-django/pull/860#discussion_r1531969853

Currently Django 5 is released, since you are already working on it, can this field be override with a system environment, and make it something like:
`if django.VERSION[:2] == (4, 2) or os.env.get('USING_DJANGO_4', `false`) == 'true'`

django_spanner/__init__.py
jaz-la, line 51: 2.2 support is dropped actually?

django_spanner/schema.py
```diff
@@ -117,14 +117,24 @@ def create_model(self, model):
...
-                    self._create_unique_sql(model, [field.column])
-                )
+                if USING_DJANGO_4:
+                    self.deferred_sql.append(
+                        self._create_unique_sql(model, [field])
+                    )
+                else:
+                    self.deferred_sql.append(
+                        self._create_unique_sql(model, [field.column])
+                    )
```
jaz-la, line 127: feel it should be:
```
if USING_DJANGO_3:
  xxx
else:
  yyy
```
as future django version will be more likely compatible with 4

django_spanner/__init__.py
```diff
@@ -34,22 +33,24 @@
...
+    SmallAutoField,
+    BigAutoField,
+)
+from django.db.models import JSONField
 
 __version__ = pkg_resources.get_distribution("django-google-spanner").version
 
 USE_EMULATOR = os.getenv("SPANNER_EMULATOR_HOST") is not None
 
 # Only active LTS django versions (2.2.*, 3.2.*) are supported by this library right now.
```
jaz-la, line 50: the comment need to be updated accordingly

django_spanner/__init__.py
```diff
@@ -34,22 +33,24 @@
...
-SUPPORTED_DJANGO_VERSIONS = [(2, 2), (3, 2)]
+SUPPORTED_DJANGO_VERSIONS = [(2, 2), (3, 2), (4, 2)]
 
 check_django_compatability(SUPPORTED_DJANGO_VERSIONS)
```
jaz-la, line 53: can you help added a:
```
bypass_django_version_check = str(os.env.get('BYPASS_DJANGO_VERSION_CHECK', 'false')) == 'true'
if not bypass_django_version_check:
  check_django_compatability(SUPPORTED_DJANGO_VERSIONS)
```
so users can test django 5 with our own risk

jaz-la:
thanks a lot for working on it, left a few comments.
I'm hoping that we can also have a way to bypass the version check and run Django 4 logic with Django 5 at our own risk.
Another suggestion is to add a summary in the PR description so new developers can understand the difference between django 3 vs 4

django_spanner/compiler.py
ankiaga: We are decoupling the code for different Django versions so that if we need to make any changes for Django 4 it should be easy to do. Also I dont see it easy to have common code as there are significant differences in 3/4

django_spanner/__init__.py
ankiaga: Sorry we cant do that as we are not sure how will Django 5 behave with this code. 
Please feel free to create a fork and make changes

django_spanner/schema.py
ankiaga: Thanks for pointing out. Made changes throughout

tests/unit/django_spanner/test_compiler.py
```diff
@@ -81,14 +103,24 @@ def test_get_combinator_sql_difference_all_sql_generated(self):
...
+            )
+        elif USING_DJANGO_4:
+            self.assertEqual(
+                sql_compiled,
+                [
+                    "SELECT tests_number.num AS col1 FROM tests_number WHERE "
+                    + "tests_number.num <= %s EXCEPT ALL SELECT tests_number.num "
+                    + "AS col1 FROM tests_number WHERE tests_number.num >= %s"
+                ],
+            )
```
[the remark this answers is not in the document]
ankiaga: No, why?

django_spanner/__init__.py
ankiaga: Thanks, done

django_spanner/__init__.py
ankiaga: As mentioned earlier we cant do that. 
Please feel free to create a fork and make changes

README.rst
```diff
@@ -140,6 +140,19 @@ configured:
            }
        }
 
+  Transaction support in autocommit mode
+  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+  We have started supporting transactions in autocommit mode by default for V4.2
+  while they are not supported in prior versions (i.e. V3.2).
```
olavloite, line 147:
```suggestion
  Django version 4.2 and higher by default supports transactions in autocommit mode.
  A transaction is automatically started if you define an
  [atomic block](https://docs.djangoproject.com/en/4.2/topics/db/transactions/#controlling-transactions-explicitly).
  
  Django version 3.2 and earlier did not support transactions in autocommit mode with Spanner.
  You can enable transactions in autocommit mode with Spanner with the
  `ALLOW_TRANSACTIONS_IN_AUTO_COMMIT` configuration option.
```

README.rst
```diff
@@ -140,6 +140,19 @@ configured:
...
+
+  - To override supporting transactions in autocommit mode in V3.2 please enable
+  the flag "ALLOW_TRANSACTIONS_IN_AUTO_COMMIT" with value True in your
+  settings.py file.
+  - To override disabling transactions in autocommit mode in V4.2 please disable
+  the flag "ALLOW_TRANSACTIONS_IN_AUTO_COMMIT" with value False in your
+  settings.py file.
```
olavloite, line 154:
```suggestion
  - To disable transactions in autocommit mode in V4.2, set
  the flag "ALLOW_TRANSACTIONS_IN_AUTO_COMMIT" to False in your
  settings.py file.
```

django_spanner/__init__.py
olavloite: To provide a bit more context here: While it would be clear to everyone involved in this thread that such an option would be 'at your own risk and not supported', it might cause surprises for other people who discover this option and start using it. Especially if Django 5.x makes changes in the future that break something in combination with this library. As it currently has zero tests covering Django 5.x, it would not be noticed by anyone, and could cause failures in production.

django_spanner/base.py
```diff
@@ -123,6 +124,15 @@ def instance(self):
             project=os.environ["GOOGLE_CLOUD_PROJECT"]
         ).instance(self.settings_dict["INSTANCE"])
 
+    @property
+    def allow_transactions_in_auto_commit(self):
+        if "ALLOW_TRANSACTIONS_IN_AUTO_COMMIT" in self.settings_dict:
+            return self.settings_dict["ALLOW_TRANSACTIONS_IN_AUTO_COMMIT"]
+        if USING_DJANGO_3:
+            return False
+        if USING_DJANGO_4:
```
olavloite, line 133: nit: remove this if statement and just return true. That makes this work by default if we in the future support Django 5 or higher.

django_spanner/base.py
```diff
@@ -205,15 +215,12 @@ def is_usable(self):
...
         Start a transaction explicitly in autocommit mode.
-
-        Staying in autocommit mode works around a bug that breaks
-        save points when autocommit is disabled by django.
         """
-        self.connection.cursor().execute("SELECT 1")
+        if self.allow_transactions_in_auto_commit:
+            self.connection.cursor().execute("BEGIN")
+        else:
+            # This won't start a transaction and was a bug in Spanner Django 3.2 version.
```
olavloite, line 225:
```suggestion
            # This won't start a transaction and was a bug in Spanner Django 3.2 version.
            # Set ALLOW_TRANSACTIONS_IN_AUTO_COMMIT = True in your settings.py file to enable
            # transactions in autocommit mode for Django 3.2.
```

django_spanner/compiler.py
olavloite: Could we add a comment to this method that explains __why__ we are overriding it, and highlight the line(s) that are actually different from the core Django implementation (so add an explicit comment where it is different)? That makes it easier to understand this, and easier the next time that a major version upgrade is being done.

django_test_suite_4.2.sh
```diff
@@ -56,6 +56,7 @@ DATABASES = {
        'NAME': "$TEST_DBNAME_OTHER",
    },
 }
+USE_TZ = False
```
olavloite, line 59: What is the default in Django 4.2? The reason for asking is that if the default in Django 4.2 is `USE_TZ = True` and we need to override it here to make all tests pass, then users should also override this in their settings. That again should clearly be called out in the release notes.
Or is there any way that we can automatically override this default in our Django implementation?

noxfile.py
```diff
@@ -26,7 +26,7 @@
 
 DEFAULT_PYTHON_VERSION = "3.8"
 SYSTEM_TEST_PYTHON_VERSIONS = ["3.8"]
-UNIT_TEST_PYTHON_VERSIONS = ["3.6", "3.7", "3.8", "3.9"]
+UNIT_TEST_PYTHON_VERSIONS = ["3.8", "3.9", "3.10"]
```
olavloite, line 29: Should we call out in the release notes that Python 3.6 and 3.7 are no longer supported?

tests/unit/django_spanner/test_operations.py
```diff
@@ -111,67 +113,162 @@ def test_convert_uuidfield_value_none(self):
...
-        self.assertEqual(
-            self.db_operations.date_extract_sql("week", "dummy_field"),
-            "EXTRACT(isoweek FROM dummy_field)",
-        )
+        if USING_DJANGO_3:
+            self.assertEqual(
+                self.db_operations.date_extract_sql("week", "dummy_field"),
+                "EXTRACT(isoweek FROM dummy_field)",
+            )
+        elif USING_DJANGO_4:
```
olavloite, line 121: Here also change to `else` blocks to (hopefully) simplify the next major version bump.

django_spanner/__init__.py
grjones: Can I second @jaz-la's request? If we can appropriately set expectations with the env car, what's the harm? I am currently using a fork that is hard to maintain with Django 5, and this PR includes everything that I have needed for Django 5 so far. Please? 

Thanks again for all of your work.

django_spanner/__init__.py
jaz-la: @olavloite I don't have a strong preference here, your point makes sense too.
We can make a folk and override this part of code to build our own Django 5 compatible version which is fine IMO.

django_spanner/compiler.py
jaz-la: @ankiaga there are a few examples:
1. looks like L53 and L127 can be consolidated
2. seems that L43~L63 can be replaced by same code in L120~L144 as `compiler.query` is always there in both Django 3/4
3. seems that L64~L118 can be replaced with Django 4 logic too, there are only 2 real differences at L80(vs L162) and L105(vs L195)

I don't have my IDE setup to run unit tests, but by taking a glance, these real differences can be put in multiple smaller if/else blocks with the majority common code out of the big if/else

django_test_suite_4.2.sh
ankiaga: Default is False in 4.2 and it would be changed to True in 5.0 https://docs.djangoproject.com/en/4.2/topics/i18n/timezones/

We need to put it explicitly here as tests were failing otherwise with a warning

noxfile.py
ankiaga: Sure thanks, let me add

django_spanner/compiler.py
ankiaga: @olavloite Added relevant comments in the code

@jaz-la Thanks for the explanation but as mentioned we don't want to optimize the code for different versions and keep it decoupled unless they are very straight forward .

django_spanner/base.py
ankiaga: Done

tests/unit/django_spanner/test_operations.py
ankiaga: Done

django_spanner/compiler.py
jaz-la: Repeating the code violates the [DRY principle](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself), here likely >90% of the code in the if/else are the same, it has potential issues:
1.it's error prone and a maintenance burden, whatever change needed in the common code need to be made twice in the future
2. in the future if we need to support Django 5.2, this part will become a tech debt to clean up if django 5.2 has another minor change needed

As long as it works it's fine for me as a user, will defer it to other reviewers.

olavloite: The reason that we are reluctant to combine these two is exactly for maintenance reasons:
1. Django 3.2 will be deprecated, and is therefore extremely unlikely to receive any further updates. That means that this code block for Django 3.2 is also extremely unlikely to require any changes in the future.
2. It is conceivable that Django 4.2 will change things related to this code block, that might require us to change things here specifically for Django 4.2. Having a shared block with Django 3.2 would only complicate such changes.
3. For a future upgrade to Django 5.x, it is potentially a lot easier if we have a separate code block for Django 4.x, as that means that during such an upgrade we can consider whether having a shared block for Django 4.x and Django 5.x makes sense. Having a code block that is shared between Django 3.2 and 4.2 would make further sharing with Django 5.x even harder.

jaz-la (approved):
thank you for working on it, I have been wanting to upgrade my django version for a long time and this is the last library that blocks me from doing so.

olavloite (approved):

PR #961, opened 2026-03-13
fix: db_returning should only be overridden for Spanner

olavloite:
The db_returning attribute of AutoFields should only be overridden for Spanner database connections, and not globally for all databases.

Fixes #https://github.com/googleapis/google-cloud-python/issues/15930


sakthivelmanii (approved):

chalmerlowe:
For clarity, the `do-not-merge` label is due to the fact that steps are being undertaken to migrate this repo to the mono-repo, likely this afternoon.
As soon as that finishes, this PR can be ported over to the mono-repo.

